PR #1, opened 2014-07-04
7.0 product variant refactor

sebastienbeau:
It's a total refactor of product variant multi.
The idea is too split the module in order to be more flexible.

As you can see a new concept have been added "Product display", this are the equivalent of "Configurable Product" on Magento side.

We do not map the "Product Template" with "Configurable Product" because it's not the same concept. Template are here for generating all of the variant, and the "Product Display" is just a way to show your product in a catalog.

For example, you can have 1 template (t-shirt color: red/yellow/blue size:S/M/L/XL) and then on magento you want to create 3 configurables products, one for each color. this case can be supported as you will create 3 displays from the product template.


product_variant_display_generator/product_variant_display.py
```diff
@@ -0,0 +1,130 @@
...
+            'product_tmpl_id',
+            domain=[
+                ('is_display', '=', False),
+                '|',
+                ('active', '=', True),
+                ('active', '=', False),
+                ], string='Product Variants'),
+    }
+
+    def _get_combinaison(self, cr, uid, product_temp, context=None):
```
bealdav, line 60: _get_combinaison seems to be franglais not english

bealdav, line 60: product_template could be abreviate product_tmpl not product_temp. It seems to be clearer and closest than odoo code. _temp could be seen as temporary

PR #2, opened 2014-07-21
[FIX] inherited form does not have Lots

kingofsevens:
When running gives XML error.. this fixes and places the page item..

### DIFF
--- a/product_variant_multi_advanced/product.xml
+++ b/product_variant_multi_advanced/product.xml
@@ -15,7 +15,7 @@
             <field name="inherit_id" ref="product_variant_multi.product_variant_form_view" />
             <field name="arch" type="xml">
                 <data>
-                    <page string="Lots" position="after">
+                    <page string="Prices" position="after">
                         <page string="Descriptions">
                             <separator string="Sale description" colspan="4"/>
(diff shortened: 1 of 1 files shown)


rvalyi:
:-1:   reject

Hello @kingofsevens 

If I'm not wrong, the variant modules will need some ground rework in v8/master after OpenERP SA improved the variant management in the core but without considering very well OCA developments unfortunately.

I need to talk in more details with the latest projects maintainers such as @sebastienbeau but I think we will just move the module to an 'unported' folder on master or v8 to explicit the module were not ported yet and then later we would rework them for V8/master.

Therefore I current reject your merge proposal, I hope you understand. If you want to contribute, then the steps are the ones I mentioned I believe.

What do you think.

kingofsevens:
Well, no offense taken. We have recently used these modules in a project. I wanted to do a quick fix. This is essential if someone wants to use this in v7, but I totally agree to a move to 'unported' folder since most of the work is done by v8 in the core.

rvalyi:
wait a minute, your fix is for v7?

If so please do the merge proposal against the 7.0 branch (you proposed it against "master"). Could you confirm?

kingofsevens:
Oops my bad. Yes it is against 7.0 branch.. I shall create another pull request then..

rvalyi:
that's why then. Thanks.

PR #3, opened 2014-07-21
[FIX] inherited form does not have Lots

kingofsevens:
Correct branch for the previous PR #2


rvalyi:
:+1:

I'm not quite sure why we didn't find the bug before, may be it was tested along another module lack stack and for some reason it wasn't crashing. Anyway you seems just right.

kingofsevens:
I was surprised too. I double checked but it refers to the variant form not product form so if there is no other module adding a "Lots" page to the variant form this must be it.

pedrobaeza:
:+1:

PR #14, opened 2015-10-26
[8.0][ADD] New module product_variant_cost_price

oihane:


product_variant_cost_price/README.rst
```diff
@@ -0,0 +1,58 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+====================
+Product Variant Cost
+====================
+
+This module adds a new cost field in product variant.
```
pedrobaeza: It doesn't add, but allows to use the cost field at variant level, not at product template level.

product_variant_cost_price/models/product.py
```diff
@@ -0,0 +1,74 @@
...
+            allfields=['cost_method'])['cost_method']['selection']
+
+    def default_cost_method(self):
+        default_dict = self.env['product.template'].default_get(
+            ['cost_method'])
+        return default_dict.get('cost_method')
+
+    @api.model
+    def _set_standard_price(self, product, value):
+        '''
```
pedrobaeza: Double quota for docstring

product_variant_cost_price/models/product.py
```diff
@@ -0,0 +1,74 @@
...
+        return self.env['product.template'].fields_get(
+            allfields=['cost_method'])['cost_method']['selection']
+
+    def default_cost_method(self):
+        default_dict = self.env['product.template'].default_get(
+            ['cost_method'])
+        return default_dict.get('cost_method')
+
+    @api.model
+    def _set_standard_price(self, product, value):
```
pedrobaeza: Why don't put this as a function the same that is on base, and use their inverse?

product_variant_cost_price/__init__.py
```diff
@@ -0,0 +1,16 @@
+# -*- coding: utf-8 -*-
+# (c) 2015 Ainara Galdona - AvanzOSC
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+
+from . import models
+from openerp import SUPERUSER_ID
+
+
+def load_cost_price_on_variant(cr, registry):
+    product_obj = registry['product.product']
```
pedrobaeza: Use this SQL expression for being multi-company aware:

```
        """
        INSERT INTO ir_property
            (name, type, value_float, value_integer, company_id, res_id,
             fields_id)
        SELECT
            ir_property.name,
            ir_property.type,
            ir_property.value_float,
            ir_property.value_integer,
            ir_property.company_id,
            'product.product,' || product_product.id::text,
            ir_model_fields.id
        FROM
            ir_property,
            product_product,
            ir_model_fields
        WHERE
            ir_property.name IN ('standard_price', 'cost_method')
        AND
            substring(ir_property.res_id from 18)::integer =
                product_product.product_tmpl_id
        AND
            ir_model_fields.model = 'product.product'
        AND
            ir_model_fields.name = ir_property.name
        ;
        """
```

The strange thing is that I haven't found any ir_property entry for cost_method. Can you investigate about it?

pedrobaeza:
This module should hide also the fields from product template form. Another option is to propagate changed values to children, but this is less intuitive for users when they see a different value in template against product.

product_variant_cost_price/__init__.py
oihane: To be able to configure 'cost_method' there had to be configured in Warehouse configuration, by default it is disabled.
So I tested... I have enabled it and then look into the db looking for the cost_method, but only if I change one of them it creates the ir_property

pedrobaeza:
Tests are still missing.

oihane:
I know, I didn't finish them but I did not want to lose this changes.

pedrobaeza:
OK, I label it WIP

rafaelbn:
If anyone is using in production environment this module https://github.com/odoomrp/odoomrp-wip/tree/8.0/product_variant_cost which names product_variant_cost and now the name is product_variant_cost_price will have problems. Isn't it? Maybe the name change must be done for v9

pedrobaeza:
No, no problem. Rename module with openupgradelib. Migration scripts are also out of scope.

oihane:
@rafaelbn althought it cames from the same module, they are not the same. All modules with product_variant_cost won't work properly with this one

rafaelbn:
@oihane product_variant_cost_price is not the same module like product_variant_cost? This PR is not a migration to OCA from a module of OdooMRP?

pedrobaeza:
@rafaelbn, it's another approach for the same feature/need. This is the valid one.

oihane:
@rafaelbn @pedrobaeza as the approach has changed, there is a field that has disappeared, so modules using product_variant_cost for that field must be fixed.

rafaelbn:
cc @JMODaERP please test, thanks

rafaelbn:
Please @JMODaERP test this one.

```
+Updates the product variant's cost when product template cost is updated.
+Adds a new inventory field on quants, (product variant cost * quant quantity).
+New proceeding on the installation of the module to load the product template's
+cost in the new field of product variant.
```

cc @JavierIniesta

product_variant_cost_price/__init__.py
```diff
@@ -0,0 +1,37 @@
+# -*- coding: utf-8 -*-
+# (c) 2015 Ainara Galdona - AvanzOSC
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+
+from . import models
+from openerp import SUPERUSER_ID
+
+
+def load_cost_price_on_variant(cr, registry):
```
nhomar: Can you explain why is this SQL needed?

product_variant_cost_price/models/product.py
```diff
@@ -0,0 +1,74 @@
...
+            for product in self:
+                product._set_standard_price(product, values['standard_price'])
+        return super(ProductProduct, self).write(values)
+
+
+class ProductPriceHistory(models.Model):
+    _inherit = 'product.price.history'
+
+    product_id = fields.Many2one(
+        comodel_name='product.product', string='Product', ondelete='cascade')
```
nhomar: Why is this field needed?

product_variant_cost_price/__init__.py
pedrobaeza: This is for putting the template price in all the product variants.

product_variant_cost_price/models/product.py
pedrobaeza: This is to have the product variant granularity in the price history. Before this module: history per template. After this module: history per variant.

product_variant_cost_price/models/product.py
```diff
@@ -0,0 +1,74 @@
...
+    standard_price = fields.Float(
+        string='Cost Price', digits=dp.get_precision('Product Price'),
+        help="Cost price of the product variant used for standard "
+        "stock valuation in accounting and used as a base price on purchase "
+        "orders. Expressed in the default unit of measure of the product.",
+        groups="base.group_user", company_dependent=True)
+    cost_method = fields.Selection(
+        string="Costing Method", selection='_get_selection_cost_method',
+        help="""Standard Price: The cost price is manually updated at the end
+                of a specific period (usually every year).
```
nhomar: Can you avoid intent the help?

you can do like that:

``` python
help='Hello World'
     ' second line '
```

If not the spaces before every line are rendered in the help and it mess up the look and feel.

moylop260:
Could you create a new commit to force travis build?

rafaelbn:
ping @JMODaERP

Please @oihane could you add usage seccion explaining how to use this module, it's not clear. Thanks

product_variant_cost_price/README.rst
```diff
@@ -0,0 +1,57 @@
...
+   :alt: Odoo Community Association
+   :target: https://odoo-community.org
+
+This module is maintained by the OCA.
+
+OCA, or the Odoo Community Association, is a nonprofit organization whose
+mission is to support the collaborative development of Odoo features and
+promote its widespread use.
+
+To contribute to this module, please visit http://odoo-community.org.
```
elicoidal: https

JMODaERP:
Can you explain, what I have to test, that  expected price  have I analysing and how is calculated.

rafaelbn:
Hi, why there are nor travis and coveralls checks?

JavierIniesta:
test_default_cost_method is failing because default attribute in cost_method field is not being used.

pedrobaeza:
It's because the default is enclosed into quotes.

product_variant_cost_price/models/product.py
```diff
@@ -0,0 +1,74 @@
...
+                Average Price: The cost price is recomputed at each incoming
+                shipment and used for the product valuation.
+                Real Price: The cost price displayed is the price of the last
+                outgoing product (will be use in case of inventory loss for
+                example).""",
+        required=True, copy=True, company_dependent=True,
+        default='default_cost_method')
```
JavierIniesta: `default=default_cost_method)`

product_variant_cost_price/README.rst
```diff
@@ -0,0 +1,57 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+====================
+Product Variant Cost
+====================
+
+Updates the product variant's cost when product template cost is updated.
```
JMODaERP: This is not as you explain. When i update the cost in template product, the variants are not updated. I can not update the product variant's cost  by hand because i have not the field Costing Method.

product_variant_cost_price/models/product.py
JavierIniesta: I modified this in my local instance. This still does not work.

pedrobaeza: I see. I'm checking.

oihane: Could it be because it is a property? Did you try with a new product or with a previously created one?

pedrobaeza: Indeed, the problem is because company_dependent is True. The default method is not called in this case. I'm investigating which method is called (I think it's called company_default_get or similar).

pedrobaeza:
The defaults for property fields are done via ir.property records: https://github.com/odoo/odoo/blob/8.0/addons/stock_account/stock_account_data.xml#L17-L22

pedrobaeza:
OK, this one is ironed through unit tests and deep functional tests and it's ready to be reviewed, so go ahead!

JavierIniesta:
Tested in runbot :+1: 

Why is runbot marked as "errored"?

oihane:
Because it has some warnings, so if it is in yellow state in github is shown as errored.

pedrobaeza:
It lacked a missing ACL. Now it's solved.

JMODaERP:
I tested in runbot: O.K. :+1:

rafaelbn:
Tested in runbot :+1:

anajuaristi:
There is a bug on this module... since if the template cost price is updated, the product cost value should not be modified on the case that is different from template
+Updates the product variant's cost when product template cost is updated --> Should be added: Only in the case that both cost prices are the same

Use case:
Template cost: 100
All variants have the same: 100

Cost increases to 101 --> All variants are modified. It's correct

But... supose I fix by hand one variant to 95. 
Now... if I change to price 101 to 102, Only the ones having value 101 should be changed from template, not the one having 95

pedrobaeza:
That's not a bug, but intended as is. Whenever you change the template cost price, all variants should change that cost price to the one you have put. That is done by consistency, because all fields in standard that you change in the template are always propagated to the variants. This field shouldn't be different.

anajuaristi:
That way, it has nosense having cost field on product variant.

Template cost has nosense. The real field where cost is stored is on product. product. Each variant can have absolutely different cost, so we can not loose it. 
Template cost could be used to easily update the variant cost at starting point for instance, but once system or user is correctly calculating and setting the variant cost, it should be mantained. 

That's my point. If you don't find it right we will make the "correction" in our own branches outside OCA.

Thank you
Ana

pedrobaeza:
Of course it doesn't make sense. That's why we have created this module in first term. But what has also no sense is to give a feature that is your interpretation and has nothing in common in consistency terms with the rest of the system. I prefer to hide that value instead of having that behaviour.

anajuaristi:
If hiding the field on template could asure that template value will be not changed, that would be correct but all Odoo core processes are "attacking" that field so I'm not sure it would solve the problem.

My fear is that system could erase the values that were correctly calculated for each product. We should find a way to preserve these values.

pedrobaeza:
System always writes the value to product.product field, but as this one is the template one, it gets written in the template. Now that the field is on product, the value is written there.

PR #19, opened 2016-01-21
[8.0] Add product_variant_supplierinfo

tafaRU:
This PR contains https://github.com/serviciosbaeza/serviciosbaeza-odoo-addons/tree/8.0/product_supplierinfo_variant module just adapted to OCA conventions.

There are still missing some unit tests, so please set it as WIP.


tafaRU:
The label **work in progress** can be removed. Thanks.

sebastienbeau:
Hi, thanks for the extraction. Just a quick comment. I am not sure that making the product_id required on the product supplier is a good thing. Indeed we have a customer that have a huge number of variant. He use the same codification as his supplier and in the majority of case the price is the same for all variant. I think it's better to have the product_id option and having a computed field for getting the ssuplier (field seller_ids). If you are ok with me on this point we can merge my work in this module and remove the constraint on required on the "product_id"

sebastienbeau:
Oups link to my work https://github.com/OCA/product-attribute/pull/148/files

sebastienbeau:
@pedrobaeza what do you think? "product_id required or not"

pedrobaeza:
You can see that we populate the product_id at the beginning (and another PR is in process to avoid a nasty warning making the population before the installation), and you can see in write and create methods some checks to always assure that the product_id is filled with the correct value, so for me for sure product_id should be required.

sebastienbeau:
My problem is a functional problem.
A user can sell a product with 100 variants
- the user have exactly the same product code as the supplier code
- all variant have the same price

In my case I want to create only one supplier rule instead of 100.

What do you think?

pedrobaeza:
OK, I see your approach. For that, what we need is to:
- Restore the "Suppliers" one2many field on the template view, but filtering out the records that have product_id set.
- Add a label to explain that the Supplier list can be overridden in the product.
- It would be desirable to view in any way (readonly or similar) the suppliers list on the product view for comparing.

What do you think?

sebastienbeau:
- Restore the "Suppliers" one2many field on the template view, but filtering out the records that have product_id set.

On the product template form, I prefer to show all product supplier for all variant. Instead I think it's easier to edit the supplier here without extra click to open the variant. What do you think?
- Add a label to explain that the Supplier list can be overridden in the product.

adding some explication is really a good idea
- It would be desirable to view in any way (readonly or similar) the suppliers list on the product view for comparing.

On the product I show all supplier available including the generic one (without product_id) https://github.com/OCA/product-attribute/pull/148/files#diff-fb5f27ef56979af325fdec4367d3d282R40

pedrobaeza:
> On the product template form, I prefer to show all product supplier for all variant. Instead I think it's easier to edit the supplier here without extra click to open the variant. What do you think?

OK, having the product_id directly in that table is a good idea.

> On the product I show all supplier available including the generic one (without product_id) https://github.com/OCA/product-attribute/pull/148/files#diff-fb5f27ef56979af325fdec4367d3d282R40

But then you're preventing the direct edit of this table from product form, which is very common when you navigate through links (or click on the blue arrow to open as pop-up). Forcing the user to go to the template form is counterproductive. This can be solved having an inverse method (that deals with all the possible operations of a one2many - you know, [(0, 0, x)] and so on), for entering this data.

pedrobaeza:
@sebastienbeau, what do you think about last comments? Are you going to work on it?

sebastienbeau:
Hi sorry it seem that my last comment was not posted.
I agree with you giving the possibility to edit from the variant form is better.
I will be not able to work on it right now, but next week I have some time to spend in a flight so I can work on it. I was think about merging my work with this PR and adapting the test and propose a PR in this PR. What do you think?

bealdav:
I'm ok with Sebastien here, how can I help you @pedro & @tafaRU to make product_id not mandatory in this module ?

pedrobaeza:
@bealdav, cool! You can start from this branch and make a PR against it and @tafaRU for sure will merge it, or make another PR to have control over possible changes and superseed this one. I'll be around to review as always :smiley:

sebastienbeau:
I have proposed a PR to update this one here : https://github.com/tafaRU/product-variant/pull/4 @tafaRU can you accept it

sebastienbeau:
The new PR is here : https://github.com/OCA/product-variant/pull/32

PR #20, opened 2016-02-02
[8.0][ADD] product_variant_sale_price module

tafaRU:


product_variant_sale_price/post_init.py
```diff
@@ -0,0 +1,19 @@
...
+from openerp import SUPERUSER_ID
+
+
+def post_init_hook(cr, registry):
+    """
+    Post-install script. Because 'lst_price' has been replaced by
+    'variant_lst_price, we should be able to retrieve its values
+    and use them to fill the new fields.
+    """
+    product_obj = registry['product.product']
```
pedrobaeza, line 14: Make only one query:

```
UPDATE product_product
SET variant_lst_price = product_template.list_price
FROM product_template
WHERE product_template.id = product_product.product_tmpl_id
```

pedrobaeza:
Isn't maybe better to make the field `list_price` a regular field and remove the standard inverse computation?

eLBati:
@pedrobaeza hi.
We chose to not touch `lst_price` field in order to not remove the standard behaviour: if in some cases the standard behaviour is fine, user can use it, otherwise they use the new field `variant_lst_price`

pedrobaeza:
That's why I prefer to change it, so that you don't have to touch every process that deals with this field.

eLBati:
I mean: with the current version, if users want , for some products, to express prices based on standard `lst_price`, they can use the standard pricelist. If they want some other products to be based on `Variant Sale Price`, they can use `Variant Sale Pricelist`

Could you describe an example where you would have to touch other processes that deals with this field?

Thanks

pedrobaeza:
In a quick search, here you have a hardcoded reference to that field that is not covered by pricelists: https://github.com/odoo/odoo/blob/8.0/addons/account/account_invoice.py#L1359

eLBati:
Ok, I agree, better to touch `lst_price` trying as you suggested

product_variant_sale_price/i18n/product_variant_sale_price.pot
```diff
@@ -0,0 +1,44 @@
+# Translation of Odoo Server.
```
pedrobaeza, line 1: Remove POT file. It's not needed anymore.

product_variant_sale_price/models/product_product.py
```diff
@@ -0,0 +1,17 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openerp.osv import orm, fields
+from openerp.addons import decimal_precision as dp
+
+
+class ProductProduct(orm.Model):
+    _inherit = 'product.product'
+
+    _columns = {
```
pedrobaeza, line 12: Can you put this on new API?

pedrobaeza:
The field still remains with the old name.

pedrobaeza:
See alternative module proposition here (still a WIP): https://github.com/sergio-teruel/product-variant/blob/8.0-PR-product_variant_sale_price/product_variant_sale_price

eLBati:
Hi @pedrobaeza and @sergio-teruel  know what is the status of the module
https://github.com/sergio-teruel/product-variant/tree/8.0-PR-product_variant_sale_price/product_variant_sale_price
?

Can a PR be done?

pedrobaeza:
Yes, we'll do it next week and work over that one.

hurrinico:
@pedrobaeza i've tried to install sergio's module but it returns this error:
loading product_variant_sale_price/views/product_view.xml
Error context:
View `product.product.form.inherited`
[view_id: 1437, xml_id: product_variant_sale_price.product_normal_form_view_default_check, model: product.product, parent_id: 258]

hurrinico:
@pedrobaeza, if it's ok to you, i'm going to work on the error tomorrow.

pedrobaeza:
OK, sorry for the silence. We have been very busy these days. I'll review it when you have it.

hurrinico:
@pedrobaeza i've done this https://github.com/OCA/product-variant/pull/26 based on @sergio-teruel 's module

tafaRU:
@sergio-teruel @pedrobaeza I'm going to port https://github.com/sergio-teruel/product-variant/tree/8.0-PR-product_variant_sale_price/product_variant_sale_price to OCA and to close the current PR.
Meanwhile, could you please update me about its status?

Thank you in advance.

tafaRU:
I close this in favour of https://github.com/OCA/product-variant/pull/30

PR #23, opened 2016-04-13
Updated product_variant_multi addon to 9.0

navarromiguel:
Updated product_variant_multi addon to 9.0 with some problems on product_product model, two problems detected with _product_compute_weight_volume and _product_lst_price methods, called on product_product form view.
The rest of the module functionality is okay.


pedrobaeza:
This one can be closed as there's another approach already merged.

PR #24, opened 2016-05-02
[IMP] No duplication on re-import of prod. variants and attributes

leorochael:
While writing a product template, update an existing variant
(`product.product`) line if it has the same `default_code`, instead of
creating a new one.

Likewise, while creating an attribute line, if a line already exists for the
same attribute, update it instead of creating a new one.

### DIFF
--- a/product_variant_csv_import/product.py
+++ b/product_variant_csv_import/product.py
@@ -20,7 +20,12 @@
 #
 ##############################################################################
 
-from openerp import models, fields
+from openerp import api, fields, models
+
+
+# values taken from openerp.addons.base.ir.ir_fields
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Is this still interesting for you? There are no reviews. I can accept it if it includes a test for the new code.

leorochael:
Hi, @pedrobaeza. Sorry I haven't been able to give this more attention.

@mileo, this patch is useful to one of your clients. Do you think you can add a unit test?

It should not be hard. Try importing the same product with attributes and variants by CSV or XML twice. It should not duplicate the amount of `attribute_line_ids` or `product_variant_ids`.

Issue #22, opened 2016-04-12
product_variant_cost_price should rewrite stock_history 


mileo:
Stock History create product_id in product_price_history but stock_history report uses product_template https://github.com/odoo/odoo/blob/8.0/addons/stock_account/wizard/stock_valuation_history.py#L71

PR #25, opened 2016-05-19
Porting to Version 9.0 module product variant cost price

hurrinico:


pedrobaeza:
We should resolve https://github.com/OCA/product-variant/issues/22 in both v8 and v9

eLBati:
It seems javieria@antiun.com is not associated to any github account

pedrobaeza:
That mail is not valid anymore, but you can keep the name, because the attribution is correct. Please switch my name to "Tecnativa" in the credits, as this is my new company.

hurrinico:
Useless PR in V9

PR #27, opened 2016-06-06
[IMP] product_variant_cost_price: when template has only one variant,

oihane:
template standard_price is written twice, creating two entries in history

### DIFF
--- a/product_variant_cost_price/models/product_product.py
+++ b/product_variant_cost_price/models/product_product.py
@@ -55,6 +55,12 @@ class ProductProduct(models.Model):
 
     @api.model
     def create(self, values):
+        if 'product_tmpl_id' in values and 'standard_price' not in values:
+            template = self.env['product.template'].browse(
+                values.get('product_tmpl_id'))
+            values.update({
+                'standard_price': template.standard_price,
+            })
         product = super(ProductProduct, self).create(values)
         self._set_standard_price(product, values.get('standard_price', 0.0))
         return product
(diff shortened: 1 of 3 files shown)


pedrobaeza:
:+1:

agaldona:
LGTM :+1:

oihane:
Found more issues... creates more history than it is supposed

product_variant_cost_price/models/product_product.py
```diff
@@ -55,6 +55,12 @@ def _set_standard_price(self, product, value):
 
     @api.model
     def create(self, values):
+        if 'product_tmpl_id' in values and 'standard_price' not in values:
+            template = self.env['product.template'].browse(
+                values.get('product_tmpl_id'))
+            values.update({
```
pedrobaeza, line 61: Why do you need this?

oihane, line 61: when you create a "new" variant using, create_variants_ids method from product.template the taken price is always 0.0 not the one in the template, that way it creates 2 history lines.
and it was impacting with the price in the template, changing it to 0.0

pedrobaeza, line 61: OK, I see, so it's not a problem of the history, but the variant cost price. Thanks for clarifying.

pedrobaeza:
:+1:

oihane:
@agaldona can you take another look, please.

agaldona:
Reviewed :+1:

PR #28, opened 2016-06-06
[NEW][8.0] product_variant_weight. Allow store weights at product.product level

sergio-teruel:
This PR alow define weight and weight_net into produt.product, when updates weight in template this overwrite all variants
@Tecnativa


product_variant_weight/README.rst
```diff
@@ -0,0 +1,52 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+======================
+Product Variant Weight
+======================
+
+This module allows to have product weight at product level instead of
+product template
```
pedrobaeza: period (`.`) at the end. Include a usage section about the fact that writing these fields on the template, they propagate to the variants.

product_variant_weight/hooks.py
```diff
@@ -0,0 +1,12 @@
+# -*- coding: utf-8 -*-
+# © 2016 Sergio Teruel <sergio.teruel@tecnativa.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+
+def post_init_hook(cr, registry):
+    """This post-init-hook will update all existing product.product weight"""
+    cr.execute("""UPDATE product_product pp
```
pedrobaeza: Make 2 SQL (one per field) and put the condition `WHERE <field> is NULL` to avoid overwriting previous values (although it's not the usual, maybe you have loaded previously the values, or you have made a cycle of installing/uninstalling the module without dropping the columns).

product_variant_weight/tests/test_product_variant_weight.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def setUp(self):
+        super(TestProductVariantWeight, self).setUp()
+        self.template = self.env['product.template']
+        self.product_template = self.template.create({
+            'name': 'Product',
+            'weight': 50,
+            'weight_net': 50,
+        })
+        self.product = self.env.ref('product.product_product_4')
```
pedrobaeza: Don't use demo data, but create yours. This way, you don't depend on this data, that may vary over versions.

product_variant_weight/hooks.py
pedrobaeza: Sorry, not NULL, but 0.

product_variant_weight/tests/test_product_variant_weight.py
```diff
@@ -0,0 +1,41 @@
...
+            self.product.weight, self.product.product_tmpl_id.weight)
+        self.assertNotEqual(
+            self.product.weight_net, self.product.product_tmpl_id.weight_net)
+
+    def test_update_template_variant(self):
+        self.product.product_tmpl_id.weight = 200
+        self.product.product_tmpl_id.weight_net = 200
+        for variant in self.product.product_tmpl_id.product_variant_ids:
+            self.assertEqual(self.product.weight, variant.weight)
+            self.assertEqual(self.product.weight_net, variant.weight_net)
```
pedrobaeza: Test the post_init_hook, emptying a value in the product variant, calling the method, and checking that the value is there again (the same as in the template).

sergio-teruel:
@pedrobaeza changes done

pedrobaeza:
:+1:

pedrobaeza:
Check Travis

sergio-teruel:
@pedrobaeza changes done... I have clicked to rebuild in runbot..

pedrobaeza:
:+1: Thank you very much

product_variant_weight/models/product.py
```diff
@@ -0,0 +1,56 @@
...
+            values['weight'] = vals['weight']
+        if 'weight_net' in vals:
+            values['weight_net'] = vals['weight_net']
+        return values
+
+    @api.multi
+    def write(self, vals):
+        res = super(ProductTemplate, self).write(vals)
+        for product in self:
+            values = self._prepare_weight_values(vals)
```
yajo: Given you are doing exactly the same in `write` and `create`, I think `_prepare_weight_values` would be better as `_update_variants_weight`, and should have this logic you write from lines 24 to 26. Just for DRY.

product_variant_weight/models/product.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+class ProductProduct(models.Model):
+    _inherit = 'product.product'
+
+    weight = fields.Float(
+        string='Gross Weight', digits=dp.get_precision('Stock Weight'),
+        help='The gross weight in Kg.')
+    weight_net = fields.Float(
+        string='Weight Net', digits=dp.get_precision('Stock Weight'),
```
yajo: Net weight

product_variant_weight/models/product.py
```diff
@@ -0,0 +1,56 @@
...
+        help='The net weight in Kg.')
+
+    @api.model
+    def create(self, vals):
+        product = super(ProductProduct, self).create(vals)
+        product.write({
+            'weight': product.product_tmpl_id.weight,
```
yajo: only if `"weight" not in vals`

product_variant_weight/models/product.py
```diff
@@ -0,0 +1,56 @@
...
+            'weight_net': product.product_tmpl_id.weight_net,
```
yajo: only if `"weight_net" not in vals`

sergio-teruel:
@Yajo changes done

yajo:
Code & functional review. :+1: Kudos for having coverage at 100%! :tada:

PR #33, opened 2016-08-04
[ADD] module product_variant_search_by_attribute

bealdav:
You can search variants with string about attributes: here entry is `16` and `blac`

Odoo retrieve products where the strings  `16` and `blac` appear in the values of the attributes, whatever the positions of the strings.

ping @pedrobaeza @tafaRU @oihane , could you have a look to this PR, please ?


product_variant_search_by_attribute/README.rst
```diff
@@ -0,0 +1,83 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+=====================================
+Product Variants Search by Attributes
+=====================================
+
+Allows to search product variants according all its attributes at once:
```
elicoidal: Allows searching inside the product variant attributes:
- All attributes at once
- Multiple keyword at once

product_variant_search_by_attribute/README.rst
```diff
@@ -0,0 +1,83 @@
...
+- with only one entry in a unique field in the search box. 
+- several keywords can be entried.
+
+
+.. figure:: product_variant_search_by_attribute/src/description/1.png
+   :alt: alternative description
+   :width: 600 px
+
+
+.. figure:: src/description/2.png
```
elicoidal: add full path

product_variant_search_by_attribute/README.rst
```diff
@@ -0,0 +1,83 @@
...
+
+#. Go to product variant menu
+#. 
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/137/8.0
+
+
+Known issues / Roadmap
```
elicoidal: remove if empty

product_variant_search_by_attribute/__openerp__.py
```diff
@@ -0,0 +1,22 @@
+# coding: utf-8
+# © 2016 David BEAL @ Akretion <david.beal@akretion.com>
+# © 2016 Raphaël REVERDY @ Akretion <raphael.reverdy@akretion.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    'name': 'Product Variant Search by Attributes',
+    'summary': """Allows to search product variants matching several keywords
+                  in their attribute values.""",
+    'version': '8.0.0.0.1',
```
elicoidal: 8.0.1.0.0

product_variant_search_by_attribute/__openerp__.py
```diff
@@ -0,0 +1,22 @@
+# coding: utf-8
+# © 2016 David BEAL @ Akretion <david.beal@akretion.com>
+# © 2016 Raphaël REVERDY @ Akretion <raphael.reverdy@akretion.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    'name': 'Product Variant Search by Attributes',
+    'summary': """Allows to search product variants matching several keywords
```
elicoidal: adapt according to README

product_variant_search_by_attribute/models/attribute.py
```diff
@@ -0,0 +1,34 @@
...
+class ProductAttributeValue(models.Model):
+    _inherit = 'product.attribute.value'
+
+    @api.multi
+    def write(self, vals):
+        """ Attribute Values update impact products: we take care
+            to propagate changes on attribute_str field"""
+        res = super(ProductAttributeValue, self).write(vals)
+        if 'name' in vals:
+            sql = """SELECT prod_id
```
elicoidal, line 22: no company_id?
Can it be a problem if the product is restricted to a company not accessible by the user?

elicoidal:
small details and a question on multi-company.
Other than that, :+1:

product_variant_search_by_attribute/__openerp__.py
```diff
@@ -0,0 +1,22 @@
...
+                  in their attribute values.""",
+    'version': '8.0.0.0.1',
+    'category': 'Product',
+    'author': 'Akretion',
```
rvalyi: I think the guideline is to add the 'Odoo Community Association (OCA)' author here to give a good visibility to the OCA on that crappy appstore.

product_variant_search_by_attribute/models/attribute.py
bealdav, line 22: Attribute values have no company_id and are shared between companies. If I change an attribute value, all companies are impacted even those which have product, for which I have no access.
Then these queries update attribute_str for all products whatever companies: it's a massive update. It's the good way to do it. attribute_str is only a repetition of each attribute individually in lower string.

pedrobaeza:
What is the advantage over searching in the field "Search Attributes for:", that performs a query over the attribute many2many and works the same way?

bealdav:
@pedrobaeza good question, I'll explain in readme.
In my example, if I search '16 bla' with 'Attrib. values (search all word)', I get one record as result.
With standard 'Search Attributes for', I get no record because it search the whole string, not the split string.

My customer has ten or fifteen attributes by variant, then several thousand of variants: with this module he can with only one field search several keywords to reduce selection to ten variants or minus

pedrobaeza:
But that's not true, because it depends in the attribute order, as you are finally searching on a string, so no possibility of swapping the attribute order.

For me, this is an unnecessary module that only confuses about its possibilities, that are very tiny. :-1:

product_variant_search_by_attribute/__openerp__.py
elicoidal: Correct ;). 
Travis will complain otherwise

hparfr:
@pedrobaeza 

The actual way of searching for a product who is "blue", "dark" and "big" is hard to do by the user.
Because the order should not matter, she has to do the query in 3 times with the advanced search pane.

With this module, the search domain `[(field,"like", "blue dark big")]` is transformed to `[(field, 'like', 'blue'), (field, 'like', 'dark'), (field, 'like', 'big')]` which is what the user wants.

pedrobaeza:
Not really, because you can have an attribute called "16 GB" (like in demo data), and if you use that kind of search with the string "16 gb", you would hit other attributes like "agbar", "343416232", etc, which makes this very unpractical.

I still don't see this is the way. I prefer to make incremental searches over the existing field. It's very easy with the keyboard:
- Type one attribute.
- Move with the down arrow to the attribute search.
- Click enter.
- Repeat the process for each attribute you want to filter.

hparfr:
>   Type one attribute.
>     Move with the down arrow to the attribute search.
>     Click enter.
>     Repeat the process for each attribute you want to filter.

It will produce an "OR" not an "AND". You will end up with more results.  And that's not what the user expects : adding search terms should reduce the result set.

pedrobaeza:
Right, but not with this module and the "Shift" key pressed: https://github.com/OCA/web/tree/8.0/web_search_with_and

bealdav:
@pedrobaeza, Additionally search in product.product is very slow because of attributes display recompute one by one. 
With this module if you have 10 keywords, search takes 5 seconds instead of 30 seconds or 1 minute at least (with repeat cycling as you suggest).
Our clients have a lot of variants and have a lot of search to do with their erp.
Alternative you suggest (thanks to notice this module) is not compatible with our use cases volumetry

pedrobaeza:
OK, but please change then the split with by commas, so you can search "16 gb, white", and the searched attributes will be "16 gb" and "white".

hparfr:
- Space is already the most common separator for searchs
- It should be simple by default
- Attributes values may already contains comma
- Product db are usually not normalized (you may have "16 gb" in one place and "2GHz" in another )
- having false positive will not be an issue since the search will be on multiple terms

bealdav:
Now separator is customizable in ir.config_parameter

product_variant_search_by_attribute/models/attribute.py
elicoidal: I understand that but products do have company_id so you will need to fetch only the product from the user company or false.
your thoughts?

elicoidal, line 22: OK (I did not note at first that this is mass update!)

pedrobaeza, line 22: Anyway, why not do this through ORM? We should avoid SQL the most we are able to.

product_variant_search_by_attribute/models/product.py
```diff
@@ -0,0 +1,111 @@
...
+               order=None, context=None, count=False):
+        _logger.debug('Initial domain search %s' % domain)
+        separator = self.pool['ir.config_parameter'].get_param(
+            cr, uid, 'search.by.attribute.separator', ' ')
+        domain = self.domain_replacement(domain, 'attribute_str', separator)
+        return super(ProductProduct, self).search(
+            cr, uid, domain, offset=offset, limit=limit, order=order,
+            context=context, count=count)
+
+    def domain_replacement(self, domain, field, separator):
```
pedrobaeza, line 74: Why don't you use methods available in expression.py (expresion.OR, expresion.AND...) and this way you don't have to recreate all the domain?

product_variant_search_by_attribute/models/attribute.py
bealdav, line 22: For performance reason. If you have attribute value in 30 000 products like my customer, it takes 3 minutes with orm instead 5 seconds with sql as indicated.

product_variant_search_by_attribute/models/product.py
bealdav, line 74: expression OR is slow to be executed with any database engine, AND is really fast with any engine.

product_variant_search_by_attribute/models/attribute.py
pedrobaeza, line 22: Well, I doubt it, as is an update over a plain search, but I won't insist. Just remove the logging messages that are going to fill the log uselessly

product_variant_search_by_attribute/models/product.py
pedrobaeza, line 74: This is nothing to do with database engine, but to assure that you insert correctly the tuples in the domain. It's only a way to simplify your code.

bealdav, line 74: I'm not sure to fully understand the use case of OR expression you describe in case of complex domain, the benefit on simplicity that bring us.

product_variant_search_by_attribute/models/attribute.py
hparfr, line 22: @pedrobaeza is there a way to express this exact query with the orm ?

bealdav, line 22: @pedrobaeza I moved log to debug level

mourad-ehm:
:+1:

EBII:
efficace

jjscarafia:
:+1:

sebastienbeau:
:+1: tested ok for me

PR #34, opened 2016-08-30
[8.0][ADD] product_variant_available_in_pos module

tafaRU:


tafaRU:
Someone please set label as workinprogress.
Thank you in advance.

tafaRU:
I think PR is ready to be reviewed, so please remove label **workinprogress**.
Thank you in advance.

paolovalier:
:+1:

product_variant_available_in_pos/models/product_template.py
```diff
@@ -0,0 +1,31 @@
...
+        values = {}
+        if 'available_in_pos' in vals:
+            values['available_in_pos'] = vals['available_in_pos']
+        if values:
+            self.product_variant_ids.write(values)
+
+    @api.multi
+    def write(self, vals):
+        res = super(ProductTemplate, self).write(vals)
+        for product in self:
```
espo-tony: I believe it would be better to move the if statement actually present in the method '_update_available_in_pos' into the 'write' method just above the for statement.
`def _update_available_in_pos(self, vals):
        values = {'available_in_pos': vals['available_in_pos']}
        self.product_variant_ids.write(values)`

```
`@api.multi
def write(self, vals):
    res = super(ProductTemplate, self).write(vals)
    if 'available_in_pos' in vals:
        for product in self:
            product._update_available_in_pos(vals)
    return res`
```

This very small change wouldn't change the functionality of the involved methods but could slightly improve the performance, computing the expression in the if statement once for all instead than as many times as the records in self (notice that its value won't be changed during the iteration).

espo-tony (commented):
I think the module works fine.
I could just suggest a very little change in order to improve the performance.

coleste:
👍

product_variant_available_in_pos/models/product_template.py
tafaRU: Done at https://github.com/OCA/product-variant/pull/34/commits/c6fbdd58cf559e4fe05d42d5bab804b7ed06b122

Thanks!

espo-tony:
:+1:

eLBati:
:+1:

product_variant_available_in_pos/__openerp__.py
```diff
@@ -0,0 +1,17 @@
+# -*- coding: utf-8 -*-
+# Copyright 2016 Alex Comba - Agile Business Group
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    'name': 'Product Variant Available In Pos',
+    'summary': 'Available In Pos in product level',
+    'version': '8.0.1.0.0',
+    'category': 'Product Management',
```
pedrobaeza: Shouldn't this be Point of sale category?

product_variant_available_in_pos/README.rst
```diff
@@ -0,0 +1,53 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+================================
+Product Variant Available In Pos
+================================
+
+This module allows to have 'Available in the Point of Sale' flag at
+product level instead of product template.
```
pedrobaeza, line 10: Mention that this is incompatible with pos_product_template

tafaRU:
@pedrobaeza I've addressed your remarks. Thanks!

pedrobaeza (approved):

PR #35, opened 2016-09-02
Migrate product_variant_supplierinfo to 9.0

guewen:


elicoidal:
:+1:

pedrobaeza:
:+1: 

A Usage section would be appreciated.

sebastienbeau:
code review :+1:

product_variant_supplierinfo/models/product_supplierinfo.py
```diff
@@ -0,0 +1,33 @@
...
+# © 2015 Pedro M. Baeza <pedro.baeza@tecnativa.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openerp import models, fields, api
+
+
+class ProductSupplierInfo(models.Model):
+    _inherit = "product.supplierinfo"
+
+    product_id = fields.Many2one(
```
alexis-via, line 11: @guewen This field should have been removed from this module, because it is native starting from v9 https://github.com/odoo/odoo/blob/9.0/addons/product/product.py#L1272 and the definition here is exactly the same as in the native product module.

PR #36, opened 2016-09-21
[FIX][product_variant_weight] Avoid useless loop

tafaRU:


pedrobaeza (approved):

Garamotte (changes requested):
:+1: but you forgot the write method.

This makes the code easier to read, but doesn't really change performance, as create always works on a single record.

The same change can be done in the write method, which can be called on multiple records.

PR #38, opened 2016-10-03
[8.0][ADD] product_variant_configurator

ThomasBinsfeld:
- [x] README
- [ ] Unit tests


ThomasBinsfeld:
@pedrobaeza

pedrobaeza:
You should respect commit history following a little variation of this guide and then make the renaming: https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-9.0#technical-method-to-migrate-a-module-from-80-to-90-branch

zakiuu:
@pedrobaeza @sbidoul  Hi,
I've been trying working on the test coverage for the product_variant_configurator, and I could get it to almost 80%, Although there are some cases that can be only tested with other modules such as purchase, sales, MRP.
My target is to add the product configurator in the purchase module, where the user can select a product template, and gives values to the attributes related to the selected template in the purchase line. When saving the purchase order the 'product.product' gets created on the fly.
I've checked in 'odoomrp' repo that such module exists. My question is: 
should I port the 'odoomrp' module to OCA in this PR or create a new one?

Garamotte:
@zakiuu : Please also look at my review on the 9.0 PR : https://github.com/OCA/product-variant/pull/39#pullrequestreview-3411235
Some comments may be applicable to this PR.

zakiuu:
@sylvain-garancher Hi,
Can you please add more reviews here. I did take care of the reviews you've put in #39

product_variant_configurator/models/product_configurator.py
```diff
@@ -0,0 +1,245 @@
...
+                self.product_id = False
+            attribute_list = []
+            for attribute_line in self.product_tmpl_id.attribute_line_ids:
+                attribute_list.append({
+                    'attribute_id': attribute_line.attribute_id.id,
+                    'product_tmpl_id': self.product_tmpl_id.id,
+                    'owner_model': self._name,
+                    'owner_id': self.id,
+                })
+            self.product_attribute_ids = [(0, 0, x) for x in attribute_list]
```
Garamotte, line 57: You can directly append the tuples `(0, 0, data)` in `attribute_list` and remove the for loop.

product_variant_configurator/models/product_configurator.py
```diff
@@ -0,0 +1,245 @@
...
+                    break
+        if not self.product_id:
+            product_tmpl = self.product_tmpl_id
+            values = self.product_attribute_ids.mapped('value_id')
+            if 'partner_id' in self._fields:
+                # If our model has a partner_id field, language is got from it
+                obj = self.env['product.attribute.value'].with_context(
+                    lang=self.partner_id.lang)
+                values = obj.browse(
+                    self.product_attribute_ids.mapped('value_id').ids)
```
Garamotte, line 85: You should use `values.ids`, instead of calling the `mapped` method twice.

product_variant_configurator/models/pricelist.py
```diff
@@ -0,0 +1,161 @@
...
+            'SELECT i.id '
+            'FROM product_pricelist_item AS i '
+            'WHERE (product_tmpl_id IS NULL OR product_tmpl_id = any(%s)) '
+            'AND (product_id IS NULL) '
+            'AND ((categ_id IS NULL) OR (categ_id = any(%s))) '
+            'AND (price_version_id = %s) '
+            'ORDER BY sequence, min_quantity desc',
+            (prod_tmpl_ids, categ_ids, version.id))
+        item_ids = [x[0] for x in cr.fetchall()]
+        items = self.env['product.pricelist.item'].browse(item_ids)
```
Garamotte, line 80: Maybe we should use `items = self.env['product.pricelist.item'].browse(item_ids).exists()`, to be sure we only read available ids for the current user, because the SQL query doesn't check access rules (`ir.rule`).

@pedrobaeza : What's your opinion about that ?

pedrobaeza: This was copied directly from Odoo code, so it should be the same. That's the part that I like less, because I didn't want to override full method. Maybe we can look for a better implementation of this part.

zakiuu: @pedrobaeza @sylvain-garancher I did spend a lot of time looking at this method, even the code wasn't fully copied. I checked how it is done in v10 and have adapted to v8 but when I did unit tests I got some bugs, 
I can push it here if you want to have a look at it

pedrobaeza: Well, we can take a look.

pedrobaeza:
@zakiuu The preservation of the commit history is still needed in the PR.

Garamotte:
I just had a traceback when testing on Runbot :

```
  File "/srv/openerp/instances/openerp-oca-runbot/parts/odoo-extra/runbot/static/build/3269415-38-07fea9/openerp/addons/product_variant_configurator/models/product_configurator.py", line 45, in onchange_product_tmpl_id
    self.product_id = self.product_tmpl_id.product_variant_ids
  File "/srv/openerp/instances/openerp-oca-runbot/parts/odoo-extra/runbot/static/build/3269415-38-07fea9/openerp/fields.py", line 851, in __set__
    value = self.convert_to_cache(value, record)
  File "/srv/openerp/instances/openerp-oca-runbot/parts/odoo-extra/runbot/static/build/3269415-38-07fea9/openerp/fields.py", line 1581, in convert_to_cache
    raise ValueError("Wrong value for %s: %r" % (self, value))
ValueError: Wrong value for product.product.product_id: product.product(41, <openerp.models.NewId object at 0x7f53b9347510>)
```

Test done on this instance : https://runbot.odoo-community.org/runbot/build/3269415

Steps to reproduce : Go to the `product.product` form view, click `Create`, select a value in the `product_tmpl_id` field.

Probable cause : I was trying to create a variant for a product which already has a variant, but no attributes defined (Blank CD, in my test)

ghost:
I was also able to reproduce the error.

It occurs when creating a new Product Variant and using a product that has only 1 variant (i.e. has an empty _Variants_ tab) as template.

pedrobaeza:
@ThomasBinsfeld @zakiuu, are you going to continue this PR?

zakiuu:
Yeah ofcorse we were juste busy with a project release

pedrobaeza:
OK, thanks. Let me know any doubt.

mmalorni:
@ThomasBinsfeld @zakiuu I have someone that can work on this PR and we could complete it next week. We need to have this for a client. Do you want us to take over or do you want our help in another way. Will it be ready shortly if we just help? 
Thanks

zakiuu:
@mmalorni Any help is very welcomed.
@pedrobaeza getting back to work on the PR today

pedrobaeza:
I have extracted correctly the commit history and merge your work in https://github.com/tecnativa/product-variant/tree/8.0-product_variant_configurator. @mmalorni is going to work in this also this week. Please coordinate both.

zakiuu:
@pedrobaeza  @mmalorni  I'll continue working on increasing the test coverage and find a better implementation for the product pricelist

pedrobaeza:
OK, cool, then just replace your branch with mine and you'll have all the work correctly credited (I have squashed your work in one commit for simplicity).

product_variant_configurator/models/pricelist.py
zakiuu: @pedrobaeza what is the use of overriding the pricelist method here in this module?

pedrobaeza: The main purpose is to calculate the correct price when yo still don't have a product variant created, but the rules for product template and product categories still apply.

rvalyi: this method is more than 130 lines... I think it's a bit too much... Isn't possible to break it down into smaller pieces that would make sense for potential overriders and make it easier to read?

zakiuu: I don't believe this is the right implementation of the behavior that we want to have. 
I would take it out from this module and have a  specific module named price_list product configurator where a rule can be defined using the product attribute configurator, and give the user the option of adding the value of the extra price to the base price.
to resume:
- Add one2many product attribute configurator field to the pricelist item object.
- Add boolean field  'include extra price to base' to take care of the extra price that the variant might have.
- override the _price_rule_get_multi to include the new behavior.
  What do you think guys
  @sbidoul

pedrobaeza: This method was a copy/paste from the core one. The idea was to ease the identification of differences from upstream.  Maybe you can find a way to inherit correctly without doing this hard copy. @oihane can comment about the problems she found.

rvalyi: @pedrobaeza got it. In this case indeed this is acceptable indeed. Such a comment in the code would be welcome.

lmignon:
@pedrobaeza Thank you for your help! I try to clarify in my head what you expect with this PR. If I check the diff between this branch and yours, I can see that the module `purchase_product_variant` is not part of you branch. Am I right? Do you expect a replace of the content of the branch referenced in this pr with the one you provide and an other PR with `purchase_product_variant`?

lmignon:
@pedrobaeza reset done on your branch!

pedrobaeza:
Hi, @lmignon. Yeah, the idea is to finish this module in one PR (for not delaying more their inclusion), and then make another PR on top of this one with the purchase_product_configurator module. This way, we respect the implicit rule of 1 module = 1 PR, and it's also easier to follow commit history, as joining both is going to be a very long list (although I squashed a bit the history).

What do you think?

lmignon:
@pedrobaeza As you can see I've reset this branch on yours. In the same time I prepare an other one with the work done on `purchase_product_configurator` based on this one and keeping the commit history from odoomrp

pedrobaeza:
Cool! Let me know if you need help with anything, although I know that you're a master git-fu :wink:

lmignon:
@pedrobaeza split and history recovery done for `purchase_product_variants` in #43

zakiuu:
Yeah any help is welcome

zakiuu:
@pedrobaeza I've fixed the issue in the pricelist I think the branch can be merged now

ghost:
@zakiuu hey there - do you still need help regarding the test coverage for this PR (or is it actually ready)? If so, I may have some time to work on these, just tell me where I should focus.

zakiuu:
@sc-microcom it will be really good if you add some the test coverage is at 86%

mmalorni:
On the variant tab view "product.template.product.form" of the product form, there is a message that says that modifying the list of variants attributes will delete variants and loose data. Maybe this text should be reviewed since variants are not created automatically in this module.
"adding or deleting attributes
will delete and recreate existing variants and lead
to the loss of their possible customizations."
Replace by:
"Variants will not be created automatically. When created, they will be limited to the attributes and values entered here"

mmalorni:
The form view "product.product.form" doesn't work. When you want to create a variant, it displays the old Attribute values drop down and then the grid doesn't work. It is not used since when you only put a value in the grid for one of the attributes it give a warning that the product already exists.

mmalorni:
Looks like the rename of the views and model files in commit 58dd25 has not removed the views and models they were replacing. We now have 10 views instead of 5. Some are identical duplicates but some have differences. Which one is the right one?

zakiuu:
I did fix that

mmalorni:
@zakiuu Maybe the manipulation that @lmignon or @pedrobaeza did to get back the commit history and to split in two the PR to keep the purchase part separate caused those changes to be lost but in this branch the files are back. I see that in the original branch (8.0-add_product_variant_configurator_original) I assume, the files are not there as you say but this is not the branch that is in the PR.

zakiuu:
I'll push it later on

mmalorni:
@zakiuu Hi, do you need help to fix the files issue? Do you know what could have caused the problem? Is it one of your commit that is missing? 
We will go ahead and add tests and do a PR on your branch if it is ok. You can then merge it to add the tests.

zakiuu:
I've just pushed the commit fix

pedrobaeza:
Is this working then?

zakiuu:
@pedrobaeza it is a yes for me

pedrobaeza:
@mmalorni, do you confirm also that your issues are solved?

mmalorni:
@pedrobaeza I will test today (5h later than you guys). We started working on the test coverage but we will be delayed. My dev called sick today.

Sent from my iPad
Martin Malorni
@:mmalorni@videotron.ca
@:Martin.malorni@hec.ca
H:(450) 962-9842
M:(514) 831-7037

mmalorni:
@pedrobaeza @zakiuu , the removal of the duplicate forms didn't resolve the issues. My comments stands. The drop down is still not removed and the grid doesn't work.

mmalorni:
@pedrobaeza Do we really need to get the coverage to 96% ? This seems high for code coverage considering that some parts cannot be tested. Is there à max % where the scripts stops asking for more coverage?
We added some tests on some files and got the % to 82% but it will be hard to get this file to 96%.

pedrobaeza:
Well, the more we have, the better, but if you find hard times doing that, the important thing is to get this working and merge it, and continue improving on later versions if possible.

mmalorni:
@pedrobaeza @zakiuu @lmignon We have done a first PR on the acsone branch with the additional tests. Can someone merge those tests to see what the result is in the OCA PR? We will work on fixing the code tomorrow to make the module functional.

sbidoul:
@mmalorni thanks for the effort with the tests. I'm personally quite happy with the test coverage. Of course we come from almost 100% in an almost empty repo, so being at 80% is good enough for me.

Note we are currently putting all our efforts on the 9.0 branch in #44. In particular we have significantly reworked then onchanges (in particular in https://github.com/OCA/product-variant/pull/44/commits/8594867917526bc5bd7798f945ddc37f31435416) so the attribute list remains consistent whatever the order in which the product template and/or product are changed.

So if you absolutely need to work on 8.0, I can happily merge your PR in here. Otherwise if you can work on 9.0 it's best to do your PR on top of #44 and when everything is clean there (we're almost done), backport the changes to 8.0, it will be less efforts.

mmalorni:
@Sbidoul We actually do not need the v8, we just didn¹t know that the work
on the v9 was progressing independently and faster than on the v8. We
actually need the v9 version which we were planning on tackling after. We
will help you finish PR #44 and see if we can backport in v8.

mmalorni:
We found that the module product_variant_csv_import in v8 is preventing the module from working. The issue is not apparent yet in v9 because the import module has not been converted yet.

sbidoul:
Closing. If there is interest for a 8.0 version, better to backport from 9.0.

PR #41, opened 2016-10-18
[FIX][product_variant_supplierinfo] _get_product_supplier method for use in multicompany

chafique-delli:


product_variant_supplierinfo/models/procurement.py
```diff
@@ -15,14 +15,17 @@ def _get_product_supplier(self, procurement):
         """returns the main supplier of the procurement's product
            given as argument"""
         product = procurement.product_id
+        seller = product.suspend_security().seller_id
```
pedrobaeza: You don't depend on module `base_suspend_security`. Why `sudo()` is not enough?

product_variant_supplierinfo/models/procurement.py
```diff
@@ -15,14 +15,17 @@ def _get_product_supplier(self, procurement):
...
             ('product_id', '=', product.id),
             ('company_id', '=', procurement.company_id.id),
-            ], limit=1)
+        ], limit=1)
         if company_supplier:
             return company_supplier.name
-        return procurement.product_id.seller_id
+        elif (seller.company_id == procurement.company_id or
+                not seller.company_id):
+            return procurement.product_id.seller_id
```
pedrobaeza, line 34: Why this strange fallback?

product_variant_supplierinfo/models/procurement.py
chafique-delli: @pedrobaeza , indeed, the sudo() is sufficient in this case.

product_variant_supplierinfo/models/procurement.py
chafique-delli: @pedrobaeza , In multi-companies, if the user does not belong to the company's main product supplier, then it's crashing because the user is not allowed to access the supplier's infos (you will get an error pop-up window)

pedrobaeza: OK, I see, thanks for the explanation. A comment in code in the `elif` would be good for future references.

pedrobaeza (approved):

chafique-delli:
@pedrobaeza , After discussion with @sebastienbeau , it is best not to change the module but to manage this problem of multi-company with the ir rules.
So i close the P.R.

PR #42, opened 2016-10-21
[ADD] product_variant_sale_delay module

tafaRU:


eLBati (approved):

pedrobaeza (approved):

pedrobaeza:
I'm wondering if isn't more interesting to have only one module that adds all missing fields in product variants?

tafaRU:
> I'm wondering if isn't more interesting to have only one module that adds all missing fields in product variants?

@pedrobaeza,
I think it should be a good idea to avoid further efforts, the only disadvantage I can see is forcing to use all these fields at product level.

Any suggestion about its techical name?

pedrobaeza:
product_variant_template_data can be a possibility

tafaRU:
@pedrobaeza, considering that this module is already installed in production environment of our customer and we started the migration to v10.0, we would like to keep the module as is, in order to avoid further efforts.

Thanks.

tafaRU:
@pedrobaeza, as was decided during the Barcelona Code Sprint, this PR can be merged while the version 10 will be moved to a new one module called `product_variant_template_data`.
So, please merge it!

Thank you in advance.

lasley:
Merging - disclaimer, I didn't actually look at any of this but there were enough reviews & no other commits after them

PR #44, opened 2016-10-28
[9.0][MIG] purchase product variant configurator

zakiuu:
Migration of product_variant_configurator and purchase_product_variant_configurator modules


chrisdec:
Functional tests : 
- When I create a new product, I have this error message : "Changing this parameter may cause automatic variants creation' --> This warning message must only appear if I change the variant creation field.
- Maybe it could be nice to add a menu "product variants" in purchase. I installed the sale module to have access to variants. 
-  I have a error message when I add a product without attribute to a purchase order. ( to reproduce : create a new product without attribute, create a supplier info, save and then, add this product to a purchase order). The error message is the following : 

### Odoo Server Error

  File "/srv/openerp/instances/openerp-oca-runbot/parts/odoo-extra/runbot/static/build/3270933-44-d50d01/openerp/http.py", line 648, in _handle_exception
    return super(JsonRequest, self)._handle_exception(exception)
  File "/srv/openerp/instances/openerp-oca-runbot/parts/odoo-extra/runbot/static/build/3270933-44-d50d01/openerp/http.py", line 685, in dispatch
    result = self._call_function(*_self.params)
  File "/srv/openerp/instances/openerp-oca-runbot/parts/odoo-extra/runbot/static/build/3270933-44-d50d01/openerp/http.py", line 321, in _call_function
    return checked_call(self.db, *args, *_kwargs)
  File "/srv/openerp/instances/openerp-oca-runbot/parts/odoo-extra/runbot/static/build/3270933-44-d50d01/openerp/service/model.py", line 118, in wrapper
    return f(dbname, _args, *_kwargs)
  File "/srv/openerp/instances/openerp-oca-runbot/parts/odoo-extra/runbot/static/build/3270933-44-d50d01/openerp/http.py", line 314, in checked_call
    result = self.endpoint(_a, *_kw)
  File "/srv/openerp/instances/openerp-oca-runbot/parts/odoo-extra/runbot/static/build/3270933-44-d50d01/openerp/http.py", line 964, in __call__
    return self.method(_args, *_kw)
  File "/srv/openerp/instances/openerp-oca-runbot/parts/odoo-extra/runbot/static/build/3270933-44-d50d01/openerp/http.py", line 514, in response_wrap
    response = f(_args, *_kw)
  File "/srv/openerp/instances/openerp-oca-runbot/parts/odoo-extra/runbot/static/build/3270933-44-d50d01/openerp/addons/web/controllers/main.py", line 889, in call_kw
    return self._call_kw(model, method, args, kwargs)
  File "/srv/openerp/instances/openerp-oca-runbot/parts/odoo-extra/runbot/static/build/3270933-44-d50d01/openerp/addons/web/controllers/main.py", line 881, in _call_kw
    return getattr(request.registry.get(model), method)(request.cr, request.uid, _args, *_kwargs)
  File "/srv/openerp/instances/openerp-oca-runbot/parts/odoo-extra/runbot/static/build/3270933-44-d50d01/openerp/api.py", line 250, in wrapper
    return old_api(self, _args, *_kwargs)
  File "/srv/openerp/instances/openerp-oca-runbot/parts/odoo-extra/runbot/static/build/3270933-44-d50d01/openerp/api.py", line 381, in old_api
    result = method(recs, _args, *_kwargs)
  File "/srv/openerp/instances/openerp-oca-runbot/parts/odoo-extra/runbot/static/build/3270933-44-d50d01/openerp/models.py", line 6070, in onchange
    record._onchange_eval(name, field_onchange[name], result)
  File "/srv/openerp/instances/openerp-oca-runbot/parts/odoo-extra/runbot/static/build/3270933-44-d50d01/openerp/models.py", line 5927, in _onchange_eval
    method_res = method(self)
  File "/srv/openerp/instances/openerp-oca-runbot/parts/odoo-extra/runbot/static/build/3270933-44-d50d01/openerp/addons/purchase_product_variant_configurator/models/purchase_order_line.py", line 38, in onchange_product_tmpl_id
    res = super(PurchaseOrderLine, self).onchange_product_tmpl_id()
  File "/srv/openerp/instances/openerp-oca-runbot/parts/odoo-extra/runbot/static/build/3270933-44-d50d01/openerp/addons/product_variant_configurator/models/product_configurator.py", line 49, in onchange_product_tmpl_id
    self.product_tmpl_id.product_variant_ids[0].id
  File "/srv/openerp/instances/openerp-oca-runbot/parts/odoo-extra/runbot/static/build/3270933-44-d50d01/openerp/models.py", line 5760, in __getitem__
    return self._browse(self.env, (self._ids[key],))
IndexError: tuple index out of range 
- Price on the supplier info is not taken into account when you work with attributes. (I tested the template supplier info and the variant supplier info, my price is always "0")
- The last attribute you encoded on the purchase order line is not taken into account. For example, I put a "veste" with color white and size "S" but only color white is registered. My created product variant is a "veste withe" and not a "veste white size s". I think it's a cursor problem. 

So, I stop here and I will restart tests on Wednesday :) Thanks @zakiuu !

sbidoul:
@zakiuu there are a few lint warnings in travis.
https://travis-ci.org/OCA/product-variant/jobs/171589902#L295

product_variant_configurator/models/product_category.py
```diff
@@ -0,0 +1,36 @@
...
+        default=True)
+
+    @api.onchange('no_create_variants')
+    def onchange_no_create_variants(self):
+        if not self.no_create_variants:
+            return {'warning': {
+                'title': _('Change warning!'),
+                'message': _('Changing this parameter may cause'
+                             ' automatic variants creation')
+            }}
```
sbidoul, line 26: This warning is always raised when opening the product creation form. We need to be smarter when deciding to raise the warning or not.

purchase_product_variant_configurator/models/purchase_order_line.py
```diff
@@ -0,0 +1,70 @@
...
+
+    @api.onchange('product_tmpl_id')
+    def onchange_product_tmpl_id(self):
+        if not self.product_tmpl_id:
+            return {}
+
+        result = super(PurchaseOrderLine, self).onchange_product_tmpl_id()
+
+        self.date_planned = datetime.today().strftime(
+            DEFAULT_SERVER_DATETIME_FORMAT)
```
sbidoul, line 27: Use fields.Datetime.now(). 

That said, is it normal to use now() for date_planned?

sbidoul (changes requested):

purchase_product_variant_configurator/models/purchase_order_line.py
zakiuu: This is taken from the base

product_variant_configurator/models/product_configurator.py
```diff
@@ -0,0 +1,254 @@
...
+                # If our model has a partner_id field, language is got from it
+                product = self.env['product.product'].with_context(
+                    lang=self.partner_id.lang).browse(self.product_id.id)
+            self.product_attribute_ids = [(0, 0, x) for x in attribute_list]
+            self.product_tmpl_id = product.product_tmpl_id.id
+            self.name = self._get_product_description(
+                product.product_tmpl_id, product, product.attribute_value_ids)
+
+    @api.multi
+    def onchange_product_id_product_configurator_old_api(
```
sbidoul: I guess this method can be removed in v9?

pedrobaeza: As this is a "framework", I would let it until v10 where old API is removed, although it has not been used in sale nor purchase modules.

product_variant_configurator/models/product_configurator.py
```diff
@@ -0,0 +1,254 @@
...
+            return {}
+
+        # First, empty current list
+        self.product_attribute_ids = [
+            (2, x.id) for x in self.product_attribute_ids]
+        if not self.product_tmpl_id.attribute_line_ids:
+            self.product_id = \
+                self.product_tmpl_id.product_variant_ids[0].id
+        else:
+            if not self.env.context.get('not_reset_product'):
```
sbidoul: I could not find where not_reset_product is set. @zakiuu @pedrobaeza do you know what it's for?

zakiuu:
Same haven't seen it in the code but maybe used in other modules on odoomrp
repo

product_variant_configurator/models/product_configurator.py
```diff
@@ -0,0 +1,254 @@
...
+    def unlink(self):
+        """Mimic `ondelete="cascade"`."""
+        attributes = self.mapped("product_attribute_ids")
+        result = super(ProductConfigurator, self).unlink()
+        if result:
+            attributes.unlink()
+        return result
+
+    @api.model
+    def check_configuration_validity(self, vals):
```
sbidoul: @zakiuu this one is unused? Was it used in the original code?

product_variant_configurator/models/product_configurator_attribute.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+class ProductConfiguratorAttribute(models.Model):
+    _name = 'product.configurator.attribute'
+
+    owner_model = fields.Char(required=True)
+
+    owner_id = fields.Integer(string="Owner",
+                              required=True,
+                              ondelete="cascade")
```
sbidoul: Does it make sense to have ondelete="cascade" on an Integer field?

zakiuu:
It's a reference to an abstract model, I didn't like it me too so I used a
many2many to not have to use it but Pedro showed me some examples where
that worked. So I've reverted it

sbidoul:
@zakiuu @pedrobaeza I finally found a workaround (https://github.com/OCA/product-variant/pull/44/commits/748b3a6958e655395f0366e204b27cab0223f2a1) for the error with possible_value_ids so we can keep the computed field. Pfeew... not easy to find, and it's an Odoo web client bug IMHO.

Anyway, progress there is.

@zakiuu can you look after the failing test and the remaining comments (I'll add a few more questions)?

product_variant_configurator/models/product_configurator_attribute.py
sbidoul: @zakiuu I understand why it is an Integer. What I don't understand is the ondelete=cascade. It has no effect, IMO.

purchase_product_variant_configurator/__openerp__.py
```diff
@@ -0,0 +1,35 @@
...
+              "Serv. Tecnol. Avanzados - Pedro M. Baeza,"
+              "ACSONE SA/NV,"
+              "Odoo Community Association (OCA)",
+    "contributors": [
+        "Oihane Crucelaegui <oihanecrucelaegi@avanzosc.es>",
+        "David Díaz <d.diazp@gmail.com>",
+        "Pedro M. Baeza <pedro.baeza@serviciosbaeza.com>",
+        "Ana Juaristi <ajuaristio@gmail.com>",
+        "Zakaria Makrelouf(Acsone) <z.makrelouf@gmail.com>",
+    ],
```
sbidoul, line 26: Contributors must go in the readme.

sbidoul (changes requested):

product_variant_configurator/models/product_configurator.py
pedrobaeza: It's used here for avoiding a problem in the stock move: https://github.com/odoomrp/odoomrp-wip/blob/0bdffaa4be5e5607421b17cb1ad2df12e483f336/stock_product_variants/models/stock_move.py#L50

sbidoul:
I made some changes in https://github.com/OCA/product-variant/pull/44/commits/458e27aee3511686c4b0e30272ee2e0b6cdf67fb so the code is easier to compare with #38.

More important, in https://github.com/OCA/product-variant/pull/44/commits/8594867917526bc5bd7798f945ddc37f31435416 I did a big refactoring of the onchanges so the attributes manipulation is located in reusable and easy(er) to understand methods. And I restructered the product_tmpl and product onchanges so the attribute list always remain consistent, whatever the order in which the user changes the product/product template.

Now it feels much more robust and less surprising. Try it on the purchase order.

product_variant_configurator/models/product_configurator.py
sbidoul: @pedrobaeza do you remember where this check_configuration_validity is supposed to be used?

product_variant_configurator/models/product_configurator.py
```diff
@@ -0,0 +1,272 @@
...
+        :type vals: dict
+        :raises: exceptions.ValidationError: If the check is not valid.
+        """
+        if any(not x[2].get('value_id') for
+               x in vals.get('product_attribute_ids', [])):
+            raise exceptions.ValidationError(
+                _("You have to fill all the attributes values."))
+
+    @api.multi
+    def _create_variant_from_vals(self, vals):
```
sbidoul, line 212: @pedrobaeza this method _create_variant_from_vals is used nowhere. Do you remember it's purpose?

purchase_product_variant_configurator/models/purchase_order_line.py
```diff
@@ -0,0 +1,56 @@
...
+            self.name += '\n' + self.product_tmpl_id.description_purchase
+
+        fpos = self.order_id.fiscal_position_id
+        company_id = self.env.user.company_id.id
+        self.taxes_id = fpos.map_tax(
+            self.product_id.supplier_taxes_id.filtered(
+                lambda r: r.company_id.id == company_id)
+        )
+        self._suggest_quantity()
+        self._onchange_quantity()
```
sbidoul: @zakiuu these methods _suggest_quantity and _onchange_quantity do nothing if self.product_id is not set. That's why the quantity and price is not proposed.

Instead of replicating all this product-related logic for the product template, I think it's better to create the product earlier (as soon as all the attribute values are set - we can create a product in an onchange, right?)  and let the normal logic do it's work on the newly created product.

The only drawback (compared to creating the product upon purchase order confirmation) is that we may end up creating some variants that are not used because the user has changed some attribute values, but it is not a real problem in my opinion. And the logic will be so much simpler.

@pedrobaeza what do you think?

cc/ @chrisdec

product_variant_configurator/models/product_configurator.py
Garamotte: @sbidoul This method is used in the stock_product_variants module, in odoomrp-wip.

sbidoul: @pedrobaeza @sylvain-garancher would you agree to removed unused code from here and reintroduce it when needed?

Garamotte: I have no idea about this : I just ran a `git grep _create_variant_from_vals` in the odoomrp-wip repository, and don't really know these modules :)

purchase_product_variant_configurator/views/inherited_purchase_order_views.xml
```diff
@@ -0,0 +1,55 @@
...
+                <xpath expr="//field[@name='order_line']/tree//field[@name='price_subtotal']"
+                       position="after">
+                    <field name="state" invisible="1"/>
+                    <button name="action_duplicate"
+                            state="draft"
+                            string="Duplicate"
+                            type="object"
+                            icon="STOCK_COPY"
+                            attrs="{'invisible': [('state', 'not in', ['draft', 'sent'])]}"
+                            help="Duplicate line"/>
```
sbidoul: @zakiuu can you remove this duplicate button. IMO it's unrelated to this module.

purchase_product_variant_configurator/models/purchase_order_line.py
pedrobaeza: I prefer to not mess the concept of not creating product variants until the confirmation of the order. It's simple to implement the missing logic of that methods for the template, so better to go to that part.

product_variant_configurator/models/product_configurator.py
pedrobaeza: It's used the same for stock_product_variants, as we can't create variants the regular way. Maybe we can use convert_to_write and convert_to_read methods for using a standard way.

product_variant_configurator/models/product_configurator.py
pedrobaeza: No, you can remove it. It's a trace of old code.

purchase_product_variant_configurator/models/purchase_order_line.py
sbidoul: @pedrobaeza we the immediate creation in https://github.com/OCA/product-variant/pull/44/commits/645a1e5de9ca70b1a69638b5d6de5ba611f94ba4. It's easy and works very well. It is also much less surprising for the user. Finally preserving the existing logic based on product.product is not only much simpler, but allows to interoperate with other modules that would override it. 

I see many advantages, and only a small drawback. Can you please reconsider?

pedrobaeza: Your solution doesn't work when you use _product_variant_types_ (to be renamed) and you have optional attributes, and you will get a lot of intermediate crap products. It also happens when you wrongly select the last attribute and change to another, that will continue creating a lot of non valid product.product. Even more, if you make a lot of quotations with possible variants, but they never reach the confirmed status, you will have to deal with all these created variants.

I prefer to stay as is and deal instead with the proper changes on methods. Other option is to create the product on memory, call upstream onchanges, and dispose it later, but this is very heavy for the performance.

ghost:
the method _get_po_line_values_from_proc in procurement_order is not called anywhere while processing the purchase_order. it's copied I think from 8.0 and it's not useful in this case. if we want to add product template to purchase order we should do it elsewhere.

product_variant_configurator/models/product_configurator.py
```diff
@@ -36,28 +36,58 @@ def _compute_price_extra(self):
...
+            attribute_lines = self.product_attribute_ids.browse([])
+            for vals in self.product_id._get_product_attributes_values_dict():
+                vals['product_tmpl_id'] = self.product_id.product_tmpl_id
+                vals['owner_model'] = self._name
+                vals['owner_id'] = self.id
+                attribute_lines += attribute_lines.new(vals)
+            self.product_attribute_ids = attribute_lines
+
+    def _empty_attributes(self):
+        self.product_attribute_ids = self.product_attribute_ids.browse([])
```
zakiuu, line 62: @sbidoul we should delete those unneeded records.

purchase_product_variant_configurator/views/inherited_purchase_order_views.xml
```diff
@@ -9,38 +9,29 @@
             <field name="priority" eval="20" />
             <field name="arch" type="xml">
                 <xpath expr="//field[@name='order_line']/tree" position="attributes">
+                    <!-- force using the form -->
                     <attribute name="editable"/>
                 </xpath>
-                <xpath expr="//field[@name='order_line']/tree//field[@name='product_id']"
-                       position="before">
-                    <field name="product_tmpl_id" />
-                </xpath>
```
sbidoul: @zakiuu product_tmpl_id must stay visible in the tree

purchase_product_variant_configurator/views/inherited_purchase_order_line_views.xml
```diff
@@ -14,20 +14,6 @@
...
-        </record>
-
-        <record model="ir.ui.view" id="purchase_order_line_extended_tree_view">
-            <field name="name">purchase.order.line.extended.tree</field>
-            <field name="model">purchase.order.line</field>
-            <field name="inherit_id" ref="purchase.purchase_order_line_tree" />
-            <field name="arch" type="xml">
-                <field name="product_id" position="before">
-                    <field name="product_tmpl_id" />
-                </field>
```
sbidoul: @zakiuu product_tmpl_id must stay visible in the tree

product_variant_configurator/models/product_product.py
```diff
@@ -0,0 +1,105 @@
...
+
+from openerp import api, exceptions, fields, models, _
+
+
+class ProductProduct(models.Model):
+    _inherit = ['product.product', 'product.configurator']
+    _name = "product.product"
+
+    # This is needed as the AbstractModel removes the delegated related field
+    name = fields.Char(related="product_tmpl_id.name")
```
Garamotte, line 15: @pedrobaeza Do you remember why the `name` field is redefined in the `product.configurator` model ?

If this is really needed, can it be possible to redefine this field using `compute` instead of `related` ?
The `related` attribute has precedence over `compute`, and it's not possible to empty it in v9 (this is now possible in v10).

Reason : I have a module that redefines the `name` field with `compute`, and don't find any other way to make both modules work together : I'd prefer to keep my custom `compute` behaviour, but even if my module depends on this one, the `related` overrides my `compute` method.

sbidoul:
The configurator on the purchase order line works fine.

It is broken on the product product form though. If someone wants to have a look at it, help is welcome.

mmalorni:
@sbidoul @zakiuu We have made 2 PR against your branch /acsone/product-variant/tree/9.0-mig_purchase_product_variant_configurator. 
It adds some tests and removes a function that is not longer necessary in v9.

product_variant_configurator/models/product_configurator.py
```diff
@@ -185,8 +185,15 @@ def unlink(self):
...
+    @api.multi
+    def check_configuration_validity(self):
+        for rec in self:
+            if any(not x.value_id for x in rec.product_attribute_ids):
+                raise exceptions.ValidationError(
+                    _("You have to fill all the attributes values."))
+
     @api.model
-    def check_configuration_validity(self, vals):
+    def check_configuration_validity_from_vals(self, vals):
```
sbidoul, line 196: @zakiuu do we need these two variants of `check_configuration_validity`? It's annoying to have duplicated code.

zakiuu: the check_configuration_validity_from_vals existed in the original code I didn't want to touch it plus it is not called anywhere, in the same time I wanted to  have a function to check the validity not from vals but for existing object (like order lines)

sbidoul: Let's remove dead code, it can be reintroduced later if needed! Actually check_configurator_validity should not be needed as there is already the same check in https://github.com/OCA/product-variant/blob/cc88b0ce71bf907c4f300ef87f0cc03e59083c2f/product_variant_configurator/models/product_product.py#L81  so please revert this commit.

purchase_product_variant_configurator/views/inherited_product_product_views.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<openerp>
+    <data>
+        <record id="variant_purchase_action" model="ir.actions.act_window">
+            <field name="name">Variantes</field>
```
sbidoul, line 5: s/Variantes/Product Variants/

product_variant_configurator/models/product_configurator.py
pedrobaeza: You have also the methods `_convert_to_write` and `_convert_to_read` that can switch from dictionary to record and vice versa to handle things in a more homogeneous way and avoid duplicated code.

mmalorni:
@sbidoul I see you have branched to another PR #45. Has work stopped on this version or is it still valid? Are the last changes requested still valid?

sbidoul:
@mmalorni we have indeed switched to that branch. We very much prefer the explicit creation of the product.product when creating the `purchase.order.line` both from a technical standpoint (the code is much simpler), compatibility with Odoo core (where everything is designed to work with product.product) and usability we showed both to functional people and the prefer #45.

I also added variants on invoice lines there and its very simple.

I think all comments here have been attended.

So I'd think we can close this PR and continue on #45 (which is stabilizing at last...). 

@pedrobaeza what's your opinion? If you have use cases where the variant creation must absolutely be delayed until SO/PO confirmation it can still be done with the exact same abstract class, while hiding the "create product now" checkbox and adding the logic to create products on PO/SO confirmation in another module.

pedrobaeza:
The only doubt I have with this approach is if you will have all the features (pricelist, minimum quantity, variant extra price, etc), if you don't create the product.product record. If not, then it's not valid for me.

sbidoul:
@pedrobaeza the logic to make purchase.order.line work with an empty product_id can be added in additional modules. All the logic in product_variant_configurator remains exactly the same. So yes it's possible and clean.

pedrobaeza:
No, but I had it in v8 and I don't want to lose it or implement it in an additional module. It must be in this one.

sbidoul:
Doing both in the same module won't work.

What's wrong with an additional module named eg `purchase_create_variant_on_confirm` or something?

pedrobaeza:
I don't understand why they are not going to work. They were working on version 8. OK about including it in an additional module, but please don't strip out this functionality.

sbidoul:
Good. I'll try to shave some time to move that from this PR to #45. This beast is consuming a huge amount of time to get right. 

Are you ok with the modules names?
- product_variant_configurator
- purchase_product_variant_configurator
- purchase_create_variant_on_confirm
- invoice_product_variant_configurator

pedrobaeza:
The last 2 seem weird:
- `purchase_create_variant_on_confirm` should be connected with its parent module. What about `purchase_product_variant_configurator_on_confirm`. It's a bit large but very tight to the parent module.
- `invoice_product_variant_configurator` why do you need this module?

pedrobaeza:
Other option is to strip `product` word for all of them and let `variant_configurator` as the main conductor. It makes sense indeed:
- product_variant_configurator.
- purchase_variant_configurator.
- purchase_variant_configurator_on_confirm.
- sale_variant_configurator.
- ...

sbidoul:
> purchase_create_variant_on_confirm should be connected with its parent module. What about purchase_product_variant_configurator_on_confirm. It's a bit large but very tight to the parent module.

I'm fine with the long name if it's not too long for Odoo.

> invoice_product_variant_configurator why do you need this module?

Customer creating invoices without PO but willing to have the right product on the invoice nevertheless.

pedrobaeza:
What about my last proposition?

Last name should be then `account_variant_configurator` or `account_invoice_variant_configurator`.

sbidoul:
Okay, purchase_variant_configurator_on_confirm is on #45

PR #45, opened 2016-11-12
[9.0] alternative purchase-product_variant_configurator

sbidoul:
This is an alternative to #44 that lets the user create the product variant himself when creating purchase order line. It has the advantage of being much more compatible with the standard Odoo flow which is based on the product instead of the product template. So there is less code to maintain and it should be more compatible with other addons enhancing purchase orders.

image


lmignon:
This branch is not yet ready to merge....
We encounter some troubles with the 'onchange' methods. For example odoo generates the following for a x2many field... [[5], [4, 11], [4, 12]]. If your inverse is configured with an ondelete='cascade', the magic command '5' (unlink all) will remove all the associated models and the command '4' (link) will fail since the items are removed....

pedrobaeza:
We already faced these problems... and they are supposed to be solved in one PR we did to Odoo. Let me look for it.

lmignon:
@pedrobaeza ok thank you....

pedrobaeza:
This one should be: https://github.com/odoo/odoo/pull/13082

lmignon:
@pedrobaeza Thank you for the link. I'll check my install

lmignon:
@pedrobaeza The same problem as the one resolved in the convert_to_write method by odoo/odoo/13082 still exists in the convert_to_onchange https://github.com/odoo/odoo/blob/9.0/openerp/fields.py#L1820 :sob: I'll make a PR in the next few days....

pedrobaeza:
I'll be around to support you.

lmignon:
@pedrobaeza Thank you. I'm overloaded with a delivery at the end of this week. I'll work on the fix the next week (at first read very easy) .

sbidoul:
@chrisdec here is the product configurator on invoice lines. Can you please test on runbot?

sbidoul:
@gurneyalex do you know why the runbot builds are marked "skipped"?

gurneyalex:
@sbidoul OCA/runbot-addons/105 should be fixed. Can you commit --amend and push -f on your pr?

product_variant_configurator/models/product_template.py
```diff
@@ -62,7 +62,7 @@ def create_variant_ids(self):
             if ((tmpl.no_create_variants == 'empty' and
                     not tmpl.categ_id.no_create_variants) or
                     tmpl.no_create_variants == 'no' or
-                    not self.attribute_line_ids):
+                    not tmpl.attribute_line_ids):
```
zakiuu, line 65: @sbidoul

sbidoul:
@gurneyalex still skipped in runbot

pedrobaeza:
OK, time to move forward. I'm going to merge the branch to continue with the rest of modules (sale_variant_configurator, types...), and this has been incubated for a long time. In case of any possible bug, we can fix it on other PRs.

I have squashed the commits to a manageable number.

Thanks to all the participants.

sbidoul:
@pedrobaeza thanks :) if you feel like it there is quite a bit of cleaning to do in the README's

pedrobaeza:
OK, I have now a hard work to do for migrating the rest of the set, but I'll do it soon.

PR #47, opened 2016-12-07
[9.0] [ADD] sale_variant_configurator and sale_stock_variant_configurator

pedrobaeza:
Modules from odoomrp-wip that add the variant configurator in sales.

@Tecnativa


DarioLodeiros:
I was testing the module and it seems to be functional, what is pending?

pedrobaeza:
Reviews :wink:

DarioLodeiros:
I think that the new product.product created from sale.order.line on confirm, does not take into account the extra price in attributes... ¿?, i think that it's a product_variant_configurator issue:  #68

pedrobaeza:
But the problem should be in base module, not here. Please try on runbot and give all the steps on the opened issue.

DarioLodeiros:
I cant run https://runbot.odoo-community.org/runbot/build/3275006
I tried a force rebuilt but after a while waiting, it gets put in skipped state... ¿?

pedrobaeza:
I have rebased over current 9.0 branch, so it may work now.

DarioLodeiros:
The module don't install
but the steps in my local docker are:
1.- activate attributes, config product with attribute x and extra price in x, and "dont create variants automatically".
2.- Create sale order with the product and attribute x... the unit price is the product price and it does not take into account the extra price in attributes
If you can fix the runbot, I go to try it

pedrobaeza:
This has been retarget to v10, which is now completed, so I close.

PR #48, opened 2016-12-15
[9.0][NEW] product_variant_standard_price_update

sergio-teruel:
New module to update cost price for all product variants

cc @tecnativa
@rafaelbn please can you review??
Thanks


rafaelbn:
Hi, I cannot test it because runbot doesn't build. It has state "skipped", maybe @pedrobaeza could help as this happended in OCA/pos similar and it was because an travis issue. In fact https://runbot.odoo-community.org/runbot/repo/github-com-oca-product-variant-137 v9 an d v10 are broken

product_variant_standard_price_update/i18n/es.po
```diff
@@ -0,0 +1,119 @@
...
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: 8bit\n"
+"Language: es_ES\n"
+"X-Generator: Poedit 1.5.4\n"
+
+#. module: product_variant_standard_price_update
+#: model:ir.ui.view,arch_db:product_variant_standard_price_update.product_template_form_view_update_standard_price
+msgid "<span>Different cost price by variants. </span>"
+msgstr "Diferente coste por variante"
```
pedrobaeza: Keep the span tags.

product_variant_standard_price_update/models/product.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+class ProductTemplate(models.Model):
+    _inherit = 'product.template'
+
+    @api.multi
+    def _compute_same_cost_variants(self):
+        for template in self:
+            cost = template.mapped('product_variant_ids.standard_price')
+            if cost.count(cost[0]) == len(cost):
```
pedrobaeza: This will fail if no variant (for example, with the product_variant_configurator module)

product_variant_standard_price_update/models/product.py
```diff
@@ -0,0 +1,51 @@
...
+        for template in self:
+            if len(template.product_variant_ids) > 1:
+                variant_prices = template.mapped(
+                    'product_variant_ids.standard_price')
+                template.standard_price = min(variant_prices)
+            else:
+                template.standard_price = \
+                    template.product_variant_ids.standard_price
+
+    same_cost_variants = fields.Boolean(compute=_compute_same_cost_variants)
```
pedrobaeza: Put a label in this field for having a nice text when doing advanced searches

pedrobaeza (changes requested):
The wizard is not needed with the new approach.

product_variant_standard_price_update/models/product.py
sergio-teruel: Zero variants???

product_variant_standard_price_update/models/product.py
sergio-teruel: This is a compute field, we can not do search

product_variant_standard_price_update/models/product.py
pedrobaeza: Exactly, this module prevents the creation of any variant until you confirm a sales/purchase order.

product_variant_standard_price_update/models/product.py
pedrobaeza: OK, you're right, but anyway, a nice label is always welcome, hehe

product_variant_standard_price_update/README.rst
```diff
@@ -0,0 +1,69 @@
...
+Product variant standard price update
+=====================================
+
+This module extends the product variant feature to allow update standard price
+in all variants from the product template form
+
+Installation
+============
+
+The module adds a button in template form for show a wizard to update the
```
pedrobaeza: This text should change if we remove the wizard (that we should).

sergio-teruel: @pedrobaeza About the wizard i have decided keep it, because it is more UX than mass editing action.

sergio-teruel:
@pedrobaeza changes done, except remove wizard :smile:

product_variant_standard_price_update/README.rst
pedrobaeza: But it's a lot of code to maintain that duplicates a function that already exists. And I don't agree it's more UX, as you have the same control and easiness with the other method.

sergio-teruel: Where is this function?? How to update all cost variant price??

pedrobaeza: You have 2 options:

* Set the price on the template, and it's propagated to all the variants, which is the most common need.
* Add the mass editing action on product.product, and change discretely the price on the variants you want.

sergio-teruel: To point one, ok, when all variant standard price is the same, but not for second point, is better that user can update this prices from a button, the mass editing option is for technical users.

pedrobaeza: I agree that we need a technical user for setting it, but not for using it, but is not commonly need, if a deployment need it, then we can add the mass editing action. Anyway, you're duplicating the variant list (in your wizard and in the smart-button variant list). I even prefer to add a static action like the mass-editing one than use that approach. Or are you going to add a wizard for every possible value that you set at variant level (weight, volume, EAN13...)?

sergio-teruel:
@pedrobaeza Changes done.

sergio-teruel:
@pedrobaeza Can you take a look?. Thanks

rafaelbn (approved):
Functionally tested 👍 many months in production... please @pedrobaeza review!

pedrobaeza:
What about my last comment? https://github.com/OCA/product-variant/pull/48#discussion_r92982671

pedrobaeza:
@sergio-teruel what happen with this PR? Can you check my comment?

sergio-teruel:
@pedrobaeza Fix readme

pedrobaeza (approved):

PR #49, opened 2017-01-07
[ADD] stock_picking_variant_mgmt

pedrobaeza:
Handle easily multiple variants on Stock Pickings
=================================================

This module allows to add/modify of all the variants of a product in a direct screen without the need of handling them one by one.

It also adds a convenient way of handling the transfer of the products in a 2D matrix with all the values of the first attribute in columns, and the rest of the combinations in rows.

Configuration
=============

* Configure your user to have any permission from "Inventory" group.
* Create a product with 1 or 2 attributes and several values.

Usage
=====

* Go to Inventory > Dashboard.
* Create a new picking from one of the existing picking types.
* Press "Add variants" button located in the upper right corner of the "Initial Demand" tab.
* A new screen will appear allowing you to select the products that have variants.
* Once you select the product, a 2D matrix will appear with the first attribute values as columns and the second one (if any) as rows.
* If there are already order lines for the product variants, the current quantity will be pre-filled in the matrix.
* Change the quantities for the variant you want and click on "Transfer to picking"
* Move lines for the variants will be created/removed to comply with the input you have done.

As extra feature for saving steps, there's also a button on each existing line that corresponds to a variant that opens the dialog directly with the product selected.

You are also able to manage variants on 1 dimension in the transfer:

* Go to the "Operations" page.
* Press on "Manage Variants Transfer" button in the upper right corner of the tab.
* Change the quantities to transfer.
* Click on "Transfer to picking" button.

Known issues / Roadmap
======================

* The inline button for modifying quantities for an existing line won't work correctly until these 2 PRs are merged in Odoo:

  * https://github.com/odoo/odoo/pull/13557
  * https://github.com/odoo/odoo/pull/13635

  The patches are already integrated on OCB.

* Make this work with product with more than 1 attribute.


pedrobaeza:
cc @Tecnativa

pedrobaeza:
@Yajo, I need your help with the JS part. There are 2 issues:

* Loading values in the 2D matrix in the default instead with an onchange, `name_get` is not resolved in many2one. I have made a workaround preloading a char field with the label.
* The change event (`xy_value_change`) is not binded correctly, and thus values are not changed. This code is not reached: https://github.com/OCA/web/blob/f20f51c3bf3342540b8ac111049de076f5eb98b8/web_widget_x2many_2d_matrix/static/src/js/web_widget_x2many_2d_matrix.js#L148

pedrobaeza:
cc @sergio-teruel

sergio-teruel:
@pedrobaeza I tested on runbot but when i try to transfer partial quantities, the line is not update.

seleccion_018

seleccion_019

yajo:
Yes @sergio-teruel, I'm checking that right now, it's what @pedrobaeza was meaning before.

sergio-teruel:
Oks. @Yajo.

yajo:
https://github.com/OCA/web/pull/518 should fix this.

pedrobaeza:
@sergio-Teruel, fixed the JS part.

yajo:
I think you'll have to rebase, to get OCA/web#518 in.

pedrobaeza:
No, a rebase is needed when you make changes in your branch, not in another. Only a rebuild was needed (and I already launched it the other day).

sergio-teruel (approved):
:+1:

stock_picking_variant_mgmt/models/stock_move.py
```diff
@@ -0,0 +1,14 @@
...
+
+from openerp import models, fields
+
+
+class StockMove(models.Model):
+    _inherit = 'stock.move'
+
+    product_attribute_value_ids = fields.Many2many(
+        comodel_name='product.attribute.value',
+        related="product_id.attribute_value_ids",
```
sergio-teruel: readonly=True
For avoid write trigger on related model

sergio-teruel:
@pedrobaeza Why this module not is in stock-logistic-workflow?
The other modules as sale_order_varint_mgmt is in sale-workflow and purchase_order_variant_mgmt is in purchase-workflow...

pedrobaeza:
I regret for putting that modules there. All of this is variant related stuff, so it should be here. My plans on v10 is to move them to this repo (we can do already the PRs, as there's no changes to make).

sergio-teruel:
@pedrobaeza Oks. Perfect...

sergio-teruel (approved):

luismontalba (approved):
Functionality tested on runbot

PR #50, opened 2017-02-13
Migrate to 10.0 mig product variant supplierinfo

mourad-ehm:


bealdav:
:+1:

product_variant_supplierinfo/models/procurement.py
```diff
@@ -15,13 +15,14 @@ def _get_product_supplier(self, procurement):
...
            given as argument"""
         product = procurement.product_id
-        company_supplier = self.env['product.supplierinfo'].search([
-            '|',
-            '&',
-            ('product_tmpl_id', '=', product.product_tmpl_id.id),
-            ('product_id', '=', False),
-            ('product_id', '=', product.id),
-            ('company_id', '=', procurement.company_id.id),
+        company_supplier = self.env['product.supplierinfo'].search(
```
pedrobaeza: Why do you change this style? Please don't make the diff bigger only for stylistic reasons (squash your commits when restoring this to avoid several diffs).

product_variant_supplierinfo/views/product_product_view.xml
```diff
@@ -7,24 +7,21 @@
...
-        <field name="inherit_id" ref="product.product_normal_form_view"/>
+        <field name="inherit_id" ref="product.product_variant_easy_edit_view"/>
         <field name="arch" type="xml">
-            <xpath expr="//notebook//field[@name='seller_ids']/preceding-sibling::separator"
-               position="attributes">
-                <attribute name="invisible">1</attribute>
+            <xpath expr="//group[@name='pricing']/parent::group"
+               position="after">
+                <separator string="Vendors"/>
+                <group name="variant_uppliers" string="Variant Vendors">
```
pedrobaeza: s/variant_uppliers/variant_suppliers

pedrobaeza (changes requested):

product_variant_supplierinfo/models/procurement.py
mourad-ehm: Hi @pedrobaeza, I changed this because I got a PEP 8 error (closing bracket does not match indentation)

pedrobaeza: But you only need to change... the closing bracket, not the rest.

mourad-ehm: OK

pedrobaeza (approved):
Please squash your 3 commits in one.

mourad-ehm:
Hi @pedrobaeza, could you merge this PR please ?

product_variant_supplierinfo/views/product_product_view.xml
```diff
@@ -7,24 +7,21 @@
     <record id="product_normal_form_view_sellers" model="ir.ui.view">
         <field name="name">product.product.form.sellers</field>
         <field name="model">product.product</field>
-        <field name="inherit_id" ref="product.product_normal_form_view"/>
+        <field name="inherit_id" ref="product.product_variant_easy_edit_view"/>
```
tafaRU, line 10: Wouldn't it be better to add new fields in both views: `product_variant_easy_edit_view` and `product_normal_form_view`?
@pedrobaeza, @mourad-ehm, what's your point of view?

product_variant_supplierinfo/models/product_supplierinfo.py
```diff
@@ -2,7 +2,7 @@
 # © 2015 Pedro M. Baeza <pedro.baeza@tecnativa.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
-from openerp import models, fields, api
+from odoo import api, fields, models
 
 
 class ProductSupplierInfo(models.Model):
```
alexis-via, line 8: The field product_id should have been removed from this module, because it is native starting from v9 https://github.com/odoo/odoo/blob/9.0/addons/product/product.py#L1272 and still native on v10 https://github.com/odoo/odoo/blob/10.0/addons/product/models/product.py#L631 and the definition here is exactly the same as in the native product module.

Issue #54, opened 2017-03-16
Product variant sale price on diferent units of measures


BorjaAlgios:
Conditions:
-The check of allow using different UOMs is marked.
-the module Product Variant Sale Price is installed.

If you select a product with units by default, but you change the UOM on the sale order to Dozens the price is multiplied by 144 instead of 12.

pedrobaeza:
Which module? Which version?

BorjaAlgios:
The version is v8.0
The module is product_variant_sale_price.

I think the problem is that in the file product_product.py of the module.
The call to 
def _compute_lst_price(self):
        for product in self:
            price = product.fix_price or product.list_price
            if 'uom' in self.env.context:
                uom = product.uos_id or product.uom_id
                **price = uom._compute_price(
                    product.uom_id.id, price, self.env.context['uom'])**
product.lst_price = price

multiplies the value by the UOM conversion, and that same conversion is applied later in the common call for the price calculation.
But, not sure if there are other implications.

pedrobaeza:
Can you make the corresponding PR and continue the discussion there?

pedrobaeza:
Closing to be tracked on the PR.

PR #55, opened 2017-03-20
[FIX] product variant sale price

BorjaAlgios:
Fix the price for the variant.
When there is a conversion with different UOM, remove the conversion
here, as it's also called outside, and the conversion is duplicated.

Issue #54


rubencabrera:
@carlosap92

pedrobaeza:
@sergio-teruel please check

pedrobaeza:
Why removing the uom part?

PR #58, opened 2017-04-28
10.0 sale order variant mgmt

pedrobaeza:
Comes from https://github.com/OCA/sale-workflow/pull/454


pedrobaeza:
Merging when finished as this has been already approved.

PR #63, opened 2017-06-28
[9.0][ADD] sale_variant_configurator_on_confirm: Migration to 9.0

DarioLodeiros:
This comes from module sale_stock_product_variants in odoomrp-wip
This PR depend on #47


pedrobaeza:
Why do you need this? The other PR is enough. The `on_confirm` was an agreement on purchase part because they stripped down the possibility of creating the variant only on confirmation, but for sale that split has not been done.

DarioLodeiros:
aiiiiissss... You are right!!
When I tried it I failed by another module that interfered
Thanks!

PR #65, opened 2017-07-04
[MIG] product_variant_configurator: Migration to 10.0

chienandalu:
Product Variant Configurator
============================

Provides an abstract model for product variant configuration.

cc @Tecnativa


product_variant_configurator/README.rst
```diff
@@ -0,0 +1,85 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+============================
+Product Variant Configurator
+============================
+
+Provides an abstract model for product variant configuration.
```
pedrobaeza: Expand the description with:
```
...configuration. It provides the basic functionality for presenting a table with the attributes of a template and the possibility to select one of the valid values. You can try this functionality creating a product variant directly selecting a product template that has attributes.

This module also prevents in a configurable way the creation of the product variants when defining the attributes and attribute values of the product template.

File: product_variant_configurator/README.rst [hunk 2]
```diff
@@ -0,0 +1,85 @@
...
+
+Configuration
+=============
+
+To configure the creation of the variants behaviour, you need to:
+
+#. Go to ``Warehouse > Products``, and select a product.
```
pedrobaeza: This menu is not available by default, but there's no menu with the base product. But it's better to say that installing `sale` module, you have to:
```
#. Go to Sales > Settings > Configuration, and select "Products can have several attributes, defining variants (Example: size, color,...)" on "Product Variants" section.
#. Go to ``Sales > Products``, and select...
```

product_variant_configurator/README.rst
```diff
@@ -0,0 +1,85 @@
...
+#. on the Variants tab edit the value of the field ``Variant Creation``.
+#. If you want to stop the automatic creation of the variant, and have the same
+   behaviour for all the products in the same category, go to ``Warehouse >
+   Products > Product Categories``, select the category and check the checkbox
+   ``Don't create variants automatically``.
+
+Usage
+=====
+
+For developer to use product configurator in your model, you need to:
```
pedrobaeza: Add the explanation for using the product configurator on the product variant:
```
#. Go to ``Sales > Sales > Product Variants``.
#. Click on "Create" button for creating a new one.
#. On the field "Product Template", select a product template that has several attributes.
#. A table with the attributes of the template will appear below.
#. Select all the attribute values and click on "Save" button.
#. A new product variant will be created for that attributes.
#. An error will raise if there's another variant with the same attribute values or if you haven't filled all the values.
```

product_variant_configurator/models/product_product.py
```diff
@@ -0,0 +1,105 @@
...
+                    return product
+        return False
+
+    @api.constrains('product_tmpl_id', 'attribute_value_ids')
+    def _check_duplicity(self):
+        for product in self:
+            domain = [('product_tmpl_id', '=', product.product_tmpl_id.id)]
+            for value in product.attribute_value_ids:
+                domain.append(('attribute_value_ids', '=', value.id))
+            other_products = self.search(domain)
```
pedrobaeza: You need to add here the context `active_test=False` for looking also on deactivated products:
```
other_products = self.with_context(active_test=False).search(domain)
```

product_variant_configurator/models/pricelist.py
```diff
@@ -0,0 +1,200 @@
...
+            :param ID uom_id: intermediate unit of measure
+        """
+        self.ensure_one()
+        if not date:
+            date = self._context.get('date', fields.Date.today())
+        if not uom_id and self._context.get('uom'):
+            uom_id = self._context['uom']
+        if uom_id:
+            # rebrowse with uom if given
+            # TODO: Avoid rewrite whole method for this sole line:
```
pedrobaeza: Look to this TODO: it seems that this line is the only one that changes (and not correctly at all as I see), so I propose to rewrite it this way:
```
    @api.multi
    def _compute_price_rule(self, products_qty_partner, date=False,
                            uom_id=False):
        """Overwrite for covering the case where templates are passed and a
        different uom is used."""
        if products_qty_partner[0][0]._name != "product.template":
            # Standard use case - Nothing to do
            return super(ProductPricelist, self)._compute_price_rule(
                products_qty_partner, date=date, uom_id=uom_id,
            )
        if not uom_id and self.env.context.get('uom'):
            uom_id = self.env.context['uom']
            ctx = self.env.context
            ctx.pop('uom')  # Remove uom context for avoiding the re-processing
            self = self.with_context(ctx)
        if uom_id:
            # rebrowse templates with uom if given
            tmpl_ids = [item[0].id for item in products_qty_partner]
            tmpl_obj = self.env['product.template']
            tmpls = tmpl_obj.with_context(uom=uom_id).browse(tmpl_ids)
            products_qty_partner = [
                (tmpls[index], data_struct[1], data_struct[2])
                for index, data_struct in enumerate(products_qty_partner)
            ]
        return super(ProductPricelist, self)._compute_price_rule(
            products_qty_partner, date=date, uom_id=False,
        )
```

pedrobaeza (changes requested):
A few lint errors:
```
product_variant_configurator/models/pricelist.py:118:37: E126 continuation line over-indented for hanging indent
product_variant_configurator/models/pricelist.py:126:33: E126 continuation line over-indented for hanging indent
```

product_variant_configurator/models/pricelist.py
chienandalu: @pedrobaeza  Yes, I put that TODO myself as the old methods are deprecated and I had to implement the new ones. I really like your method's idea but it still falls on an infinite recursion in the last return because the first element of the tuple `product_qty_partner` is going to always be a `product.template` object (the first check should call to super)

pedrobaeza: Yes, of course, call super instead in that case! It has been my fault coding this quickly. I change it accordingly.

chienandalu: Ok :smile:  Another little tweak was needed. Take a look now: https://github.com/OCA/product-variant/pull/65/commits/30ea269e64ad0745fd90c4d613ef0acae2541c5a

product_variant_configurator/models/pricelist.py
```diff
@@ -0,0 +1,52 @@
...
+        different uom is used."""
+        if products_qty_partner[0][0]._name != "product.template":
+            # Standard use case - Nothing to do
+            return super(ProductPricelist, self)._compute_price_rule(
+                products_qty_partner, date=date, uom_id=uom_id,
+            )
+        if not uom_id and self.env.context.get('uom'):
+            uom_id = self.env.context['uom']
+            ctx = dict(self.env.context)
+            ctx.pop('uom')  # Remove uom context for avoiding the re-processing
```
pedrobaeza: You can do directly (my fault): `uom_id = ctx.pop('uom')` and save the first line

pedrobaeza (approved):

chienandalu:
@pedrobaeza Done :slightly_smiling_face:

pedrobaeza:
@sbidoul can any of you review the PR as you were part of the move of this module to OCA?

product_variant_configurator/models/pricelist.py
```diff
@@ -1,214 +1,52 @@
...
+        if products_qty_partner[0][0]._name != "product.template":
+            # Standard use case - Nothing to do
+            return super(ProductPricelist, self)._compute_price_rule(
+                products_qty_partner, date=date, uom_id=uom_id,
+            )
+        if not uom_id and self.env.context.get('uom'):
+            ctx = dict(self.env.context)
+            # Remove uom context for avoiding the re-processing
+            uom_id = ctx.pop('uom')
+            self = self.with_context(ctx)
```
lmignon: @chienandalu PLZ keep `self`untouched and assign the result of the call to `with_context` to an other variable.

lmignon (changes requested):

product_variant_configurator/models/pricelist.py
chienandalu: Thanks for your review @lmignon :slightly_smiling_face: 

Well, I really need to change the context here. Would it better something like:
```
self.env.context = uom_id = ctx.pop('uom')
```

pedrobaeza: No, David, he is saying that use another variable name, like `obj`: `obj = self.with_context(ctx)`

chienandalu: Ok. How about it now?: https://github.com/OCA/product-variant/pull/65/commits/aabddabe8699fc0ec59729ddab1b5d85681bc858

gdgellatly:
as requested by @pedrobaeza 

Issues with button box on product template when no variants.
Traceability, sales button cause traceback

pedrobaeza:
@chienandalu please check what @gdgellatly says.

chienandalu:
Thanks for your review @gdgellatly I couldn't find any of the mentioned issues on runbot. Could you give me more precise steps to reproduce?

gdgellatly:
All I did was make a new template and choose don't create automatically.
Saved it then clicked all the smart buttons to see if they worked.  Those 2
gave traceback with indexerror. All you need to do is hide them I think
when no variants.

chienandalu:
@gdgellatly Seems that that's an Odoo bug. We've placed an issue though.

pedrobaeza:
Any other approval? @cubells please review then

pedrobaeza:
@chienandalu is it easy to improve coverage?

lmignon (approved):

chienandalu:
@pedrobaeza I tried but I couldn't find easy ways to do it :confused:

cubells (changes requested):
Is this behaviour correct?

Steps to reproduce:
- create a product and set "Variant creation" to "Create them automatically"
- save product.
- edit product
- add two attributes, i.e., color and memory
An error is shown: "You have to fill all the attributes values." 

Is this correct?

pedrobaeza:
@cubells That's because you are modifying a created product and thus the variant already created doesn't contain all the attributes, so when deactivating it, the error raises. @chienandalu what we can do is to see if active is False and in that case, don't make the check.

sbidoul:
@pedrobaeza relaxing constraints for inactive products is a bit unusual in Odoo, no?

pedrobaeza:
But see the case commented for my colleague @cubells:

* Create a product template without attributes. A product variant is created without attributes.
* Now you add some attributes.
* Odoo first deactivates existing variants (here is the error), and then create the new product variants with the attributes.

gdgellatly:
Yh, that's an upstream constraint and a misunderstanding of how this module
works.  I hit it before when testing and thought it was error, but actually
it's a problem with naming and documentation if anything. Automatically
just means normal. You still need to provide values.

pedrobaeza:
I don't get you, Graeme. Do you mean that we shouldn't relax that constraint. Do you hit the problem on Odoo core also?

gdgellatly:
Sure is. Attribute lines must have values. It's a required field.

pedrobaeza:
But are we talking about the same thing? @cubells can you clarify if you added values to the attributes? If not, it's Graeme issue about product template with attributes but without values in that attributes, which is correct. If you added them, then it's what I'm talking about of a constraint introduced in this module https://github.com/OCA/product-variant/pull/65/files#diff-7e7fd9fd1c61d9d758a2416fef0c4805R95

gdgellatly:
@pedrobaeza actually I just tested on runbot and it didn't error, but IMO it should.  Give me a few minutes and I'll edit this response, I don't think relaxing constraints is the right way to go here, because the constraint itself is good.

EDIT: @cubells I am assuming the issue you have experienced is one of usability.  When you selected 'Create variant automatically' did you expect to be able to just add attributes, without specifying values?

product_variant_configurator/models/product_product.py
```diff
@@ -0,0 +1,106 @@
...
+        values are set. This can be overridden to set another rules.
+
+        :raises: exceptions.ValidationError: If the check is not valid.
+        """
+        for product in self:
+            if bool(product.product_tmpl_id.attribute_line_ids.mapped(
+                    'attribute_id') -
+                    product.attribute_value_ids.mapped('attribute_id')):
+                raise exceptions.ValidationError(
+                    _("You have to fill all the attributes values."))
```
gdgellatly, line 99: @pedrobaeza Its this line here, I'm definitely in favour of keeping it, otherwise you probably get errors with your grid thing in sale order.  What I would do though is on the product variant view under attribute lines is add a required flag to attribute_value_ids, so the user gets a normal required field error.  It makes no sense to add an attribute line without attribute values.

Also I'd change this error message to something more helpful, like "Each attribute line must have values assigned for variant creation" - or something.

product_variant_configurator/models/product_configurator.py
```diff
@@ -0,0 +1,268 @@
...
+            attribute_lines = self.product_attribute_ids.browse([])
+            for vals in self.product_id._get_product_attributes_values_dict():
+                vals['product_tmpl_id'] = self.product_id.product_tmpl_id
+                vals['owner_model'] = self._name
+                vals['owner_id'] = self.id
+                attribute_lines += attribute_lines.new(vals)
+            self.product_attribute_ids = attribute_lines
+
+    def _empty_attributes(self):
+        self.product_attribute_ids = self.product_attribute_ids.browse([])
```
gdgellatly, line 90: As an aside, isn't it cleaner just to use self.env['product.configurator.attribute'] - that gives an empty recordset and more conventional.

product_variant_configurator/models/product_configurator.py
```diff
@@ -0,0 +1,268 @@
...
+                    'attribute_id': attribute_line.attribute_id.id,
+                    'product_tmpl_id': self.product_tmpl_id.id,
+                    'owner_model': self._name,
+                    'owner_id': self.id,
+                })
+            self.product_attribute_ids = attribute_lines
+
+    def _set_product_attributes(self):
+        if self.product_id:
+            attribute_lines = self.product_attribute_ids.browse([])
```
gdgellatly, line 81: As per prior comment, not blocking, but if more edits coming then could tidy.

product_variant_configurator/models/product_product.py
chienandalu: @gdgellatly Doing some some tests on the issue I see that this error only raises if you add more than one attribute-value at once. Otherwise it behaves as expected (creates variants as you add attributes and values). So I think that this check should be done in other way: https://github.com/Tecnativa/product-variant/blob/aabddabe8699fc0ec59729ddab1b5d85681bc858/product_variant_configurator/models/product_product.py#L91

cubells:
@pedrobaeza I didn't add values to the attributes.

Steps to reproduce are clear:

- create a product and set "Variant creation" to "Create them automatically"
- save product.
- edit product
- add **two** attributes, i.e., color and memory

An error is shown: "You have to fill all the attributes values."

@gdgellatly yes, it's a problem of usability. But Why I can create one product variant after save template without attributes, and two or more variants I can't?

product_variant_configurator/models/product_product.py
gdgellatly: @chienandalu I double checked that constraint, and couldn't reproduce on runbot.  I still think the constraint is correct and should be there.  However we still have the issue of the original created product.  I think what it needs is an active test.  If product.active and (rest of constraint) because otherwise the constraint is applying to deactivated products, which in the case of added attributes will make them fail this test.

pedrobaeza: So that's what I said at the beginning... @chienandalu please add that.

gdgellatly: also here is the error here 
```
if bool(product.product_tmpl_id.attribute_line_ids.mapped('attribute_id')
```
should be in pseudo code
```
if bool(product.product_tmpl_id.attribute_line_ids.filtered('attribute_value_ids' is set).mapped('attribute_id')
```
or
```
if bool(product.product_tmpl_id.attribute_line_ids.mapped('attribute_value_ids').mapped('attribute_id')
```
see the difference, one checks that the line has any attribute value ids before adding attributes to the recordset.
EDIT: clarify that it isn't a logical or, just 2 alternatives for statement

gdgellatly: @pedrobaeza I read you wrong sorry, but also note the other error here, that will need addressing too, the case as per @cubells report where the line has no attribute values.

chienandalu: I think the error is trickier than that. Debugging into it I'm seeing that on this issue cases the constraint is checked in the middle of the product's write(). As a result of this, not all the attributes are written in that moment and the check fails. Checking if product is active had no effect.

gdgellatly:
*@chienandalu *The only reason I could think is that product_tmpl_id gets
written, triggering it before the various create/update/deactivate variant
functions, but I'm only guessing, someone will knw obetter than me.  It
can't be attribute_value_ids as they are written together, not one at a
time.  What happens if you remove product_tmpl_id from the constrains().

chienandalu:
@pedrobaeza @gdgellatly @cubells Please check my last commit that hopefully makes the same check avoiding the mentioned issue.

product_variant_configurator/models/product_product.py
gdgellatly: @chienandalu I'm not sure I quite understand the constraint anymore - I thought the reason for it was that when you created a manual variant, you had to ensure completion of one value from each attribute line.  This seems to just say the attribute lines must have values, which technically they don't need to.

But I'm going to drop out of this conversation, I have surgery today and won't be back to things for a little while.

pedrobaeza: Indeed, that's not the purpose, so @chienandalu please revert to previous condition and we will check it together.

@gdgellatly have a good surgery and I wish you a soon recover.

chienandalu: @pedrobaeza Ok, I see. I dropped the commit. That being so I find it difficult to overlook the false positives for the reasons we were discussing before (https://github.com/OCA/product-variant/pull/65#discussion_r135505113) :confused:

pedrobaeza:
@luismontalba please review when Runbot finishes

product_variant_configurator/models/product_product.py
chienandalu: @gdgellatly Seen with @pedrobaeza and hopefully fixed in  https://github.com/OCA/product-variant/pull/65/commits/b99c148c1c3e8119ac17474dc6343af023302c85

luismontalba (approved):
Tested on runbot

pedrobaeza:
OK, I think this one has past a long road and some reviews, so we can merge it now and apply possible patches in successive PRs. Thanks to all reviewers.

PR #66, opened 2017-07-04
[MIG] purchase_variant_configurator: Migration to 10.0

chienandalu:
Purchase - Product variants
===========================

This module allows you to create the product variant when a purchase order is
confirmed. It adds to the purchase line a product configurator, so that
selecting a product and its attributes can be created a new product variant.

Depends on:

- [x] product_variant_configurator: https://github.com/OCA/product-variant/pull/65

cc @Tecnativa

PR #67, opened 2017-07-05
[MIG] purchase_variant_configurator_on_confirm: Migration to 10.0

chienandalu:
Purchase Variant Configurator On Confirm
========================================

Create product variants when confirming the purchase order.

Depends on:

- [x] product_variant_configurator: https://github.com/OCA/product-variant/pull/65
- [x] purchase_variant_configurator: https://github.com/OCA/product-variant/pull/66

cc @Tecnativa


purchase_variant_configurator_on_confirm/models/product_template.py
```diff
@@ -16,15 +17,14 @@ class ProductTemplate(models.Model):
     def _select_seller(self, partner_id=False, quantity=0.0,
                        date=time.strftime(DEFAULT_SERVER_DATE_FORMAT),
```
pedrobaeza: Why this strange default value? Upstream signature is `date=None`

purchase_variant_configurator_on_confirm/models/product_template.py
```diff
@@ -16,15 +17,14 @@ class ProductTemplate(models.Model):
...
                        uom_id=False):
-        # this method is copied from the standard _select_seller on the
-        # product.product model
+        """ Copy from standard: `ProductProduct._select_seller` """
         self.ensure_one()
```
pedrobaeza: Let's try a trick here, creating a "in-memory" product record and calling the method on it:
```
product = self.env['product.product'].new({
    'product_tmpl_id': self.id,
    'name': self.name,
})
return product._select_seller(
    partner_id=partner_id, quantity=quantity, date=date, uom_id=uom_id,
)
```

purchase_variant_configurator_on_confirm/models/purchase_order_line.py
```diff
@@ -44,20 +42,21 @@ def onchange_product_tmpl_id(self):
 
         fpos = self.order_id.fiscal_position_id
         company_id = self.env.user.company_id.id
-        self.taxes_id = fpos.map_tax(
-            self.product_tmpl_id.supplier_taxes_id.filtered(
-                lambda r: r.company_id.id == company_id)
-        )
+        if self.env.uid == SUPERUSER_ID:
```
pedrobaeza: Why this unfolding? It's incorrect, as other users apart from admin can have access to several companies (through a parent company)

purchase_variant_configurator_on_confirm/models/purchase_order_line.py
```diff
@@ -18,22 +19,19 @@ class PurchaseOrderLine(models.Model):
 
     @api.onchange('product_tmpl_id')
     def onchange_product_tmpl_id(self):
-        # this method is copied from the standard onchange_product_id
+        """ Copy from standard: `PurcharseOrderLine.onchange_product_id` """
```
pedrobaeza: Have you reviewed v10 standard methods and their code? Maybe you can use similar techniques to the one I suggested above.

pedrobaeza (changes requested):

chienandalu:
@pedrobaeza I've tried to implement your suggestions in the last commit. I'm dealing with this line though: https://github.com/Tecnativa/product-variant/blob/05c2db1ffc6fab320c168920a85059a8a260f6bf/purchase_variant_configurator_on_confirm/models/purchase_order_line.py#L19
I can't make self.product_id to get that context object properly although the context seems to have the right object in the instant I want to pass it. What am I doing wrong? :confused:

pedrobaeza:
I don't get what are you trying to get with that context argument

chienandalu:
I try to pass the on memory product from template copy (https://github.com/Tecnativa/product-variant/blob/05c2db1ffc6fab320c168920a85059a8a260f6bf/purchase_variant_configurator_on_confirm/models/purchase_order_line.py#L25) to the original product_id_change function

pedrobaeza (approved):
I have been testing this and fixing the found problems, so I'm going to merge after a while this module for continuing improving things

PR #75, opened 2017-07-25
[MIG] product_variant_default_code: Migration to 10

chienandalu:
Product Variant Default Code
============================

This module allows to define a reference mask on the product templates so the
variants references are automatically set. For example:

- Product template: Jacket
- Attributes:
  - Color: White, Black
  - Size: M, L
- Reference mask: `JKT01-[Color]-[Size]`

- Reference on variants:

  - `JKT01-Wh-M` Jacket White M
  - `JKT01-Bl-M` Jacket Black M
  - `JKT01-Wh-L` Jacket White L
  - `JKT01-Bl-L` Jacket Black L

cc @Tecnativa


product_variant_default_code/README.rst
```diff
@@ -0,0 +1,112 @@
...
+- Reference mask: `JKT01-[Color]-[Size]`
+
+- Reference on variants:
+
+  - `JKT01-Wh-M` Jacket White M
+  - `JKT01-Bl-M` Jacket Black M
+  - `JKT01-Wh-L` Jacket White L
+  - `JKT01-Bl-L` Jacket Black L
+
+Usage
```
pedrobaeza: Shouldn't this be `Configuration`?

product_variant_default_code/__manifest__.py
```diff
@@ -0,0 +1,26 @@
+# -*- coding: utf-8 -*-
+# Copyright 2014 AvancOsc - Alfredo de la Fuente
+# Copyright 2014 Tecnativa - Pedro Baeza
```
pedrobaeza: `Pedro M. Baeza` please

product_variant_default_code/__manifest__.py
```diff
@@ -0,0 +1,26 @@
...
+# Copyright 2014 Shine IT - Tony Gu
+# Copyright 2017 Tecnativa - David Vidal
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+{
+    'name': 'Product Variant Default Code',
+    'version': '10.0.1.0.0',
+    'author': 'OdooMRP team,'
```
pedrobaeza: Remove `OdooMRP team`

product_variant_default_code/__manifest__.py
```diff
@@ -0,0 +1,26 @@
...
+              'Avanzosc,'
```
pedrobaeza: `AvanzOSC` as it's spelled in other modules

product_variant_default_code/i18n/product_variant_default_code.pot
```diff
@@ -0,0 +1,79 @@
+# Translation of Odoo Server.
```
pedrobaeza: Remove this file.

product_variant_default_code/views/product_attribute_value_view.xml
```diff
@@ -0,0 +1,32 @@
+<?xml version="1.0"?>
+<odoo>
+
+        <!--<record id="variants_template_tree_view_inh_variantdefaultcode"-->
```
pedrobaeza: Remove this if it's not in use.

pedrobaeza (changes requested):

chienandalu:
@pedrobaeza Changes done and previous commits squashed and rebased

pedrobaeza:
Please achieve full coverage

chienandalu:
@pedrobaeza Full coverage achieved! :smile:

pedrobaeza (approved):
Excellent!

gdgellatly:
@pedrobaeza So when I port it, https://github.com/OCA/product-attribute/pull/244 it needs to be combined to match the existing module product_code_builder, I spend quite some time making them all match up and combine in one module for no one to review so I close out of frustration after 3 months.  When your team does it - insta approve - no requirement to match up?

cc @jgrandguillaume prime example of what I mail you about IRT to root causes of current OCA review issues. 
EDIT: and @pedrobaeza is one of the good guys

pedrobaeza:
@gdgellatly I didn't put any problem in the other PR. I just suggested as a quick way to reconcile approaches to merge both functionalities. 

Obviously, I don't remember about all the PRs I commented, and we have this module installed in one v8 instance that we are migrating to v10, so the migration was planned and performed by my colleague. It's approved by me as the migration itself is correct and even improved.

Now that you remind me this, we are going to rescue your PR and analyze the merge of both modules. In a first glance, I don't like too much to have a module product_code_builder_sequence, but we will review it and tell, including the code you made. If you would keep the PR opened (as we advised you several times in other PRs), we would detect it and directly take it into account for making this PR. But no, you needed to close it for whatever reason I don't get yet.

You can privately email what you want, but this is not an example. And this is why a PR process is the best way to work: for detecting any issue, including not homogeneous decisions over time.

Anyway, this is a democratic process: having enough people supporting anything, it can be approved and, for the good or the bad, be the official way. You have been very nitpicking about the comments we have put on your PRs. They are only opinions, not the absolute truth. Of course, there are rules to follow, like no overlapping features in several OCA modules, but this rule applies for the same version. In new versions, modules can change and one used in one version, to be deprecated in favor of another ones. That was why I closed your initial PR for product_variant_default_code in v8, because product_code_builder already exists in that version. 

I'm also tired to have to spend more time with you writing endless comments trying to convince you that coding the alternatives and going ahead, which is the real work and what we do, producing very good quality OCA modules.

@chienandalu please analyze the referred PR and incorporate the missing features here.

gdgellatly:
@pedrobaeza don't take wrong way - I can send you copy of mail, was only a draft of an idea.  It was about improving process, more in line with what you did with odoomrp.  The jist was that people are only motivated to review work they are immediately interested in and as such larger integrators tend to self review.

I never made a v8 PR for product_variant_default_code that I know of, I'd be surprised as I never used v8.
The sequence module I am not sure about, it is not required at all, I don't use it, I think there was an issue of compatibility, so its only there to make them compatible that they can work together.

I know the PR to which you refer, I was wrong.  I said as much at the time.

pedrobaeza:
OK, we will tell you about your PR and its reincarnation here.

msadiq1411:
Can we download ver.10 yet?

pedrobaeza:
You can always take the code in this PR (just learn how to do it reading GitHub docs).

@chienandalu, were you able to check the other module for merging functionality here?

chienandalu:
@pedrobaeza I didnt't dedicate the time yet

product_variant_default_code/views/product_view.xml
```diff
@@ -0,0 +1,34 @@
...
+            <field name="name">
+                product.template.only.form.view.inh.variantdefaultcode
+            </field>
+            <field name="model">product.template</field>
+            <field name="inherit_id"
+                   ref="product.product_template_only_form_view"/>
+            <field name="arch" type="xml">
+                <field name="default_code" position="before">
+                  <field name="reference_mask"
+                         modifiers="{'invisible':
```
gdgellatly: @chienandalu @pedrobaeza Are you sure this line works.  I had it in my own code and it didn't.  It needed

attrs={....[('attribute_line_ids', '=', [])]}

something like that anyway, attrs not modifiers, empty list not False.

chienandalu: :+1:

rafaelbn:
Hi @gdgellatly , I'm sorry if you get disappointed because after 3 months no body review you PR. This is not something you have to report by email to @jgrandguillaume ... you have to report and propose to contributor list.

3 months.... @pedrobaeza has PRs that no body review from 2015. So please ***review***. 

> prime example of what I mail you about IRT to root causes of current OCA review issues

- The only cause is that it cannot be more code than reviews. It must be more review than code. @gurneyalex  say 1 year ago, if every contributor makes 3 reviews per 1 PR proposed, OCA will work like a silk.
- here we have the example: @mkamareddine ask for download the module but he don't make a review! https://github.com/OCA/product-variant/pull/75#issuecomment-323278140 it's amazing! .
- You have 27 PR. https://github.com/pulls?utf8=%E2%9C%93&q=is%3Apr+author%3Agdgellatly . 
- I have 0 PR and hundreds of reviews.

This topic is one of the most important to overcome by OCA, but depends on the contributors and their revisions.

So please, we are all together in this great community: https://youtu.be/s0by6C2ihfg

About https://github.com/OCA/product-attribute/pull/244#issuecomment-317049761 @lasley told you that you have to work with git-aggregator.  You say:

> I can't seriously put a completely unreviewed, unmerged OCA module in production where API might change subject to review.

Yes you can, we can, OCA can

> after 3 months, the mental catchup to respond hurts my brain. I had some first reviews after 18 months the other month, couldn't even respond as I no longer even had a dev copy of that version

If you have this PR in production, for sure you can answer.

Anyway OCA in improving day by day.

My best! 😄

gdgellatly:
@rafaelbn I get it, I really do.  IRT to the review of PR after 18 months, I no longer had in production. IRT to more recent ones, all our work now is retargeting to master for a pretty quick turnaround to v11.  In fact, you can see I have been doing quite a bit of early port work for some OCA modules, so yes, now my brain is in master mode, having to think back is some catchup.

It is interesting that you respond on this PR.  A module that was originated outside of OCA for some years.  

After some months trying, I've found the process of contributing "new to the world" modules to OCA to not work for me.  The main reasons for this if I were to guess, is that by definition they are "new" so there are no interested users to review (no one has a vested interest) or there are interested users, but because of timing/communication they don't see or know about it.  I think its also the reason that larger integrators tend to instantly self-review (as they are interested),  but obviously smaller contributors, while they understand the motivations behind it, get frustrated by this behaviour.  

By way of example, I got asked to review a module recently, https://github.com/OCA/product-attribute/pull/275, which received the insta-review thumbs up with no comment within a few hours of PR by a colleague of the contributor.  No automated tests, no nothing.  The reason it was requested for me is because I had an outstanding entirely unreviewed PR with nearly identical functionality for a later version.

I try to review every module I am interested in and can provide some kind of valuable feedback.  In truth there isn't very many, maybe 1 a month I would comment on and maybe 5 a month I review, but don't feel I have anything meaningful to say.  Its also hard to know when there are PR's*.  In fact I often end up reviewing something after merge.  I think that is more the problem.

And I think @gurneyalex 's can be modified to say, if every module had 3 people interested in it, that knew about it, it would get 3 reviews, and in weeks rather than years.  It is a lack of interest rather than lack of reviewers, and if we put the focus on interesting modules (which clearly mine aren't, hence I tend to close PR as soon as that is obvious) rather than bemoaning a lack of reviewers the process would flow better.  (Although I am not against OCA paying for a technical reviewer, or enforcing review/contribution ratios in problem repos).  

So at least for me, I've found it far more satisfying for new things to just go my own way.  Its not like I am the only one who thinks this, dev repos have been discussed for years, odoomrp did same thing.  The modules are there, if they ever attract enough interested users, and they want to put in OCA, then thats fine.  We'll create a PR.

For existing modules the OCA process is far smoother and enjoyable, so I'll stick to that, and leave new stuff unitl there is a base of interested people.  I think there are some exceptions in some core repos, server tools for example where new stuff is jumped on because often it has wide applicability and is naturally attractive to OCA core community.

Anyway sorry for hijacking PR.

*Indeed, that could be a nice development, allow users to register their interest in a module, and are pinged when there is a PR on it, say on OCA website. That would definitely speed flow for existing modules.  Even for new repos, a repo maintainer could say look at a new module and think "This would be a great addition for users interested in X.  They will probably be interested in this, lets ping them". Maybe even consider review more like UAT.

rafaelbn:
So this module is all checks green with 100% code coverage ready to be tested @chienandalu ?

chienandalu:
@gdgellatly I've integrated the main major difference I found in your implementation (product attributes codes). Check it out and tell me if you miss something else that is relevant to you.

@rafaelbn Yes, now it's ready to test.

gdgellatly:
I think, and sorry I am mobile so haven't checked was the other module also
updated inactive product s which I thought was important.

chienandalu:
@gdgellatly True archived products didn't update the attribute value on their ref. Fixed in the last commit and added a test to it.

product_variant_default_code/tests/__init__.py
```diff
@@ -0,0 +1,3 @@
+# -*- encoding: utf-8 -*-
```
bealdav: s/encoding/coding

product_variant_default_code/README.rst
```diff
@@ -0,0 +1,117 @@
...
+   Product: Jacket
+   Color: Black(Bl), White(Wh) # Black and White are the attribute values;
+                                 'Bl' and 'Wh' are the corresponding codes
+   Size: L (L), XL(XL)
+   
+The automatically generated default value for the Variant reference mask will
+be `[Color]-[Size]` and so the 'default code' on the variants will be `Bl-L`,
+`Wh-L`, `Bl-XL` and `Wh-XL`.
+
+The mask value can be fully custumizedwhatever you like. You can even have
```
bealdav: s/custumizedwhatever/customizedwhatever

product_variant_default_code/models/__init__.py
```diff
@@ -0,0 +1,3 @@
+# -*- encoding: utf-8 -*-
```
bealdav: s/encoding/coding

pedrobaeza (approved):
@gdgellatly please give your final blessing.

sebastienbeau:
Hi I would like to share my experience on generating automatically the code of variant.

We did a first version on launchpad (openerp version 5) with the implementation of mask (as our customer needed it). Then for the following project my customer always ask me to automatized the generation of the "mask" because most of the time they do not want to customize manually, so most of the time I have "hack" the generation in a custom module in order to avoid adding a mask manually.
At the end, I have installed the "hack" for all of my customer and I have only the first one who was using the mask.

When we refactor the code and migrate it to 8, we decide to drop the "mask" feature in order to only use the automatic generation without mask (just always concatenate the attribute code base on the sequence of the attribute) and we share the module : https://github.com/OCA/product-attribute/tree/8.0/product_code_builder

So basically the existing module product_code_builder and product_variant_default_code do not fill really the same need. But both generate default_code.

I do not want to block this PR because
- your need it to customize each product template with a specific mask
- my need is to generate the code always with the same method based on the attribute sequences

Moreover I think that a company will use only one method (customize or not), so this depend of the need.

We plan to port the module product_code_builder with the same behaviour as 8 (and move it in this branch) but maybe we need to both rename our module and to explain that module are not compatible and user should use one of this module depending of the need.

Regarding the module name I think about

your module => product_code_builder_mask
existing module => product_code_builder_auto

OR

your module => product_variant_default_code_mask
existing module => product_code_default_code_auto

An other solution is to merge the module or to refactor but I think it will be a waste of time and the code will be more complexe for nothing.

What do you think?

gdgellatly (approved):

product_variant_default_code/models/product.py
```diff
@@ -0,0 +1,196 @@
...
+             'attribute name')
+
+    @api.model
+    def create(self, vals):
+        product = self.new(vals)
+        if not vals.get('reference_mask') and product.attribute_line_ids:
+            attribute_names = []
+            for line in product.attribute_line_ids:
+                attribute_names.append("[{}]".format(line.attribute_id.name))
+            default_mask = DEFAULT_REFERENCE_SEPARATOR.join(attribute_names)
```
gdgellatly: @pedrobaeza @chienandalu 
Firstly I think default reference seperator either needs to be an empty string or configurable (or both) - the hyphen is annoying and it was something I removed in order to bring compatibility with product_code_builder.

Second, the default mask should be default_code + ''.join([attrs]).

Do those 2 things then you have identical functionality to the original product_code_builder with a lot more flexibility.

Given its a couple mins work to bring both these modules into line, we should do it before merging IMO.  

\cc @sebastienbeau

gdgellatly (changes requested):
as per comment left

product_variant_default_code/models/product.py
gdgellatly: @pedrobaeza now I remember why I ported that sequence module.  For @sebastienbeau requirement.

I think it can be resolved by 
```


if not vals.get('reference_mask') and product.attribute_line_ids:
    default_mask = self._get_default_mask(product)

...
def _get_default_mask(self, product)
    attribute_names = []
    for line in product.attribute_line_ids:
        attribute_names.append("[{}]".format(line.attribute_id.name))
    default_mask = DEFAULT_REFERENCE_SEPARATOR.join(attribute_names)
    return default_mask
````

Or just product._get_default_mask()
What do you think?

Comment (sebastienbeau) [reply product_variant_default_code/models/product.py]: Hi, indeed, maybe it worth to merge both module now.

So we need to add the field 'prefix_code' on the product template, (the default or custom mask will only add extra char to the prefix_code to build the default_code). If the mask is empty we generate the code base on a default_function or we generate on the fly a default mask (like you have proposed)

We may add an extra group (with a option in the configuration menu) to show or not the mask on the product template

@chienandalu @pedrobaeza are you ok for merging both module

@bealdav can you propose a PR with the merge?

Comment (pedrobaeza) [reply product_variant_default_code/models/product.py]: Yes, I think we should work to have only one module. My colleague @chienandalu is already working for making this, so you owe us a drink!

Comment (bealdav) [reply product_variant_default_code/models/product.py]: +1 to merge features

**Comment by chienandalu** [comment]
Ok, guys, I just made a commit covering (or hope so) your suggestions. Check it out, please :)

**Comment by bealdav** [comment]
Seems pretty good. Thanks a lot for your reactivity. I just need to test it for real now :+1:

File: product_variant_default_code/views/sale_config_settings_view.xml [hunk 12]
```diff
@@ -0,0 +1,15 @@
...
+<odoo>
+
+    <record id="view_sales_config_default_code" model="ir.ui.view">
+        <field name="name">sale settings default code</field>
+        <field name="model">sale.config.settings</field>
+        <field name="inherit_id" ref="sale.view_sales_config"/>
+        <field name="arch" type="xml">
+            <group name="customers" position="inside">
+                <field name="group_product_default_code" widget="radio"/>
+            </group>
```
bealdav: I think it's better to have
`<group string="Products" position="inside">` instead of 'customer group because it more concerns product than partners

product_variant_default_code/views/product_view.xml
```diff
@@ -0,0 +1,39 @@
...
+                          [('attribute_line_ids', '=', [])]}"/>
+                <field name="reference_mask"
+                       attrs="{'invisible':
+                          [('attribute_line_ids', '=', [])]}"
+                       groups="product_variant_default_code.group_product_default_code"/>
+            </field>
+        </field>
+    </record>
+
+    <record id="product_normal_form_view_inh_variantdefaultcode"
```
bealdav: not blocking for me but I think it's easier if we don't re-invent xml_id each time we inherit a view.
Here it seems here `product_normal_form_view` is better.

the same for id `product_template_only_form_view_inh_variantdefaultcode` and `variants_tree_view_inh_variantdefaultcode` above

product_variant_default_code/views/product_view.xml
```diff
@@ -0,0 +1,39 @@
...
+            model="ir.ui.view">
+        <field name="name">
+            product.normal.form.view.inh.variantdefaultcode
+        </field>
```
bealdav: not blocking but in fact name in xml view is not mandatory and an automatic one is computed by framework. @ akretion we no more specify it since 3 years. => easier to read

product_variant_default_code/README.rst
```diff
@@ -0,0 +1,117 @@
...
+   :target: http://www.gnu.org/licenses/agpl
+   :alt: License: AGPL-3
+
+============================
+Product Variant Default Code
+============================
+
+This module allows to define a reference mask on the product templates so the
+variants references are automatically set. For example:
+
```
bealdav: About introducing sentence, the main feature is : automatically produce/build/generate product default_code/Reference according to attributes data in a configurable manner/behavior

May be module name could reflect more main feature ? product_variant_code_builder or any other ? => then really merged of two previous modules.

I let you reply yourself to this question, not blocking

Examples are really good

product_variant_default_code/__init__.py
```diff
@@ -0,0 +1,3 @@
+# -*- encoding: utf-8 -*-
```
bealdav: not required line because it's ascii, not blocking

product_variant_default_code/README.rst
```diff
@@ -0,0 +1,117 @@
...
+   :target: https://runbot.odoo-community.org/runbot/137/10.0
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/product_variant/issues>`_. In case of trouble, please
+check there if your issue has already been reported. If you spotted it first,
+help us smash it by providing detailed and welcomed feedback.
+
```
bealdav, line 97: Include Limitation:
`
In case of attribute name update related mask are not updated.`

It could be solved later

product_variant_default_code/models/product.py
```diff
@@ -0,0 +1,206 @@
...
+
+             'If you like, You can even have the attribute name appear more'
+             ' than once in the mask. Such as , `fancyA/[Size]~[Color]~[Size]`'
+             ' When saved, the default code on variants will be something like'
+             ' `fancyA/l~r~l` (for variant with Color "Red" and Size "L") '
+             '`fancyA/x~y~x` (for variant with Color "Yellow" and Size "XL")\n'
+
+             'Note: make sure characters "[,]" do not appear in your '
+             'attribute name')
+
```
bealdav: cutting problem
print_scr

product_variant_default_code/models/product.py
```diff
@@ -0,0 +1,206 @@
...
+def render_default_code(product, mask):
+    product.default_code = get_rendered_default_code(product, mask)
+
+
+class ProductTemplate(models.Model):
+    _inherit = 'product.template'
+
+    prefix_code = fields.Char(
+        string='Prefix code',
```
bealdav: Not blocking but friendly string should be 'Reference Prefix' because of 
internal

and the field name should be `code_prefix` in english I presume

bealdav (changes requested):
globally some details for now, I'll continue the review later

product_variant_default_code/views/product_view.xml
```diff
@@ -0,0 +1,39 @@
...
+               ref="product.product_template_only_form_view"/>
+        <field name="arch" type="xml">
+            <field name="default_code" position="after">
+                <field name="prefix_code"
+                       attrs="{'invisible':
+                          [('attribute_line_ids', '=', [])]}"/>
+                <field name="reference_mask"
+                       attrs="{'invisible':
+                          [('attribute_line_ids', '=', [])]}"
+                       groups="product_variant_default_code.group_product_default_code"/>
```
bealdav: you may add a placeholder : [attribute3][attribute1] in the field

product_variant_default_code/models/product.py
```diff
@@ -0,0 +1,206 @@
...
+            for product in products:
+                if product.reference_mask:
+                    render_default_code(product, product.reference_mask)
+        return result
+
+
+class ProductProduct(models.Model):
+    _inherit = 'product.product'
+
+    manual_code = fields.Boolean(string='Manual code')
```
bealdav: s/Manual code/Manual Reference to be homogeneous

product_variant_default_code/views/product_view.xml
```diff
@@ -0,0 +1,39 @@
...
+        </field>
+        <field name="model">product.product</field>
+        <field name="inherit_id" ref="product.product_normal_form_view"/>
+        <field name="arch" type="xml">
+            <field name="default_code" position="before">
+                <field name="manual_code" />
+            </field>
+        </field>
+    </record>
+
```
bealdav, line 30: You also should add this xml : 
```xml
    <record id="product_variant_easy_edit_view" model="ir.ui.view">
        <field name="model">product.product</field>
        <field name="inherit_id" ref="product.product_variant_easy_edit_view"/>
            <field name="default_code" position="before">
                <field name="manual_code" />
            </field>
        </field>
    </record>
```
to allow displaying this field in alternate view

product_variant_default_code/models/product.py
```diff
@@ -0,0 +1,206 @@
...
+
+    prefix_code = fields.Char(
+        string='Prefix code',
+        help='Add prefix to product variant default code',
+    )
+    reference_mask = fields.Char(
+        string='Variant reference mask',
+        help='Reference mask for building internal references of a '
+             'variant generated from this template.\n'
+
```
bealdav: Is this field support translations ? ie:
user A is english and has `Color` attribute
user B is french and has the same attribute but named `Couleur`

It seems it break the behavior:
- create a product in english with Color attribute
- switch user to french : it display correctly with 'Color' attribute in mask
- try to duplicate: "Found unrecognized attribute name in "Variant Reference Mask" because it search Couleur instead of Color
- if you create a new product now 'Color' doesn't works :-(

I think we shouldn't use a translatable field as ref in mask field: either code field (to create in product.attribute), either a new field to store the name without translate (ie 'color' english version only)

bealdav (changes requested):
Bug with translatable

product_variant_default_code/models/product.py
bealdav: @sebastienbeau could you confirm we really need prefix_code ? because reference mask can be used for that.
and it allow to be better in usability to have not this field.

bealdav:
@chienandalu 

Please could you make a review of my review. If yes, I could make a PR on your PR to contribute on my side. What do you think ?

product_variant_default_code/__manifest__.py
```diff
@@ -0,0 +1,29 @@
...
+    'version': '10.0.1.0.0',
+    'author': 'AvancOSC,'
+              'Shine IT,'
+              'Tecnativa,'
+              'Odoo Community Association (OCA)',
+    'website': 'https://odoo-community.org/',
+    'license': 'AGPL-3',
+    'category': 'Product',
+    'depends': [
+        'product',
```
bealdav, line 19: I have an error when I install module with dependency or update module `ValueError: <type 'exceptions.TypeError'>: "Model 'sale.config.settings' does not exist in registry."`

Not sure than `sale.config.settings` are in registry with only product as dependency

chienandalu:
@bealdav Sure, no problem at all in working together :smile: Your PR is welcome

bealdav:
Nice, but please just reply to all comments where you are ok (just say 'ok') and 'no' on the others and then I'll do a PR only on OK elements.

product_variant_default_code/views/sale_config_settings_view.xml
chienandalu: You can't select to a string anymore. You can position it after or before to named field in the product section where it should belong, yes

product_variant_default_code/views/product_view.xml
chienandalu: Mmm, doesnt't it?

product_variant_default_code/models/product.py
chienandalu: But oldname should be added to the field if this change is made

product_variant_default_code/models/product.py
chienandalu: Yes that's true. Anyway I leave it in your roof :slightly_smiling_face:

bealdav:
Perfect, I will make a PR next week

product_variant_default_code/models/product.py
pedrobaeza: OK for not being translatable.

product_variant_default_code/__manifest__.py
pedrobaeza: The problem here is that you are assuming that variant references are for sale. Maybe you need them for purchase only and you are forcing to install sales. Please put the configuration on the general settings wizard instead and stick to product dependency.

bealdav: Theoretically I share your vision but practically I doubt there is any company which use 'purchase' without 'sale' in the world.
And if there is one they don't use advanced 'attribute management'. But I can update my code no problem with that ;-)
Ahahah: fucking ... config.settings transient are merged in one model in v11 : thanks for that @fpodoo

mourad-ehm (approved):
Code review. Not tested

pedrobaeza:
I merge this and let any other possible comment to later PRs, because this PR has been a lot here.

PR #76, opened 2017-07-26
[8.0][FIX] product_variant_supplierinfo - Add related sequence

hparfr:
Add a supplierinfo to the product.product
Add a supplierinfo to the product.template

Current behaviour : 

Both will have a sequence = 1. The template may be choosen instead of the product during the purchase.order. 
The supplier will get incomplete or falsy RFQ.

With this fix : 

Supplierinfo on the product.product will have sequence = 5 
Supplierinfo on the product.template will have sequence = 10
It's possible to manually set a different value.

Inspiration: https://github.com/OCA/product-attribute/blob/8.0/pricelist_per_product/pricelist.py#L96


pedrobaeza:
Sorry, but I don't see the problem and I consider this an ugly patch with no clear consequences (as you are passing the same time both a value for a field and its related field).

hparfr:
>Sorry, but I don't see the problem

Ipod : 2 variants.

A supplierinfo for the template,   supplier's ref = all_ipods
img 2017-07-27 at 11 26 07

A supplierinfo for variant 16GB [A6678],  supplier's ref = ipod_16
img 2017-07-27 at 11 32 10

A supplierinfo for variant 32GB [A6679]   supplier's ref = ipod_32
img 2017-07-27 at 11 22 30

All have sequence = 1 (by default)

Result: 

img 2017-07-27 at 11 22 11


Then I create a Request for Quotation 
img 2017-07-27 at 11 23 13

We can see the supplierinfo for the variant 16GB [A6678] is all_ipods.
We were expecting ipod_16.



>and I consider this an ugly patch with no clear consequences (as you are passing the same time both a value for a field and its related field).
- This way of doing is consitent with another OCA module (price_list_per_product)
- If you have suggestions how to do this in a cleaner manner, I refactor it. :man_technologist:

pedrobaeza:
OK, thanks, now I get the problem. The solution is easiest:
```
@api.model
def create(self, vals):
    if not vals.get('product_id'):
        vals['sequence'] = vals.get('sequence', 10) * 10
    return super(ProductSupplierInfo, self).create(vals)

@api.multi
def write(self, vals):
    if vals.get('sequence', None) is None:
        return super(ProductSupplierInfo, self).write(vals)
    for suppinfo in self:
        if not vals.get('product_id', suppinfo.product_id):
            vals['sequence'] *= 10
        super(ProductSupplierInfo, suppinfo).write(vals)
    return True
```

hparfr:
Thanks @pedrobaeza it's simplier

pedrobaeza (approved):
Squashing together the commits

legalsylvain (approved):
LGTM. :+1: Code review / no test.

bealdav (approved):

bealdav:
code review

PR #37, opened 2016-09-21
[8.0] [FIX] product_variant_sale_price

tafaRU:
Please see message commits for further info.

PR #80, opened 2017-09-01
[FIX] product_variant_sale_price: avoid to update list price of all variants

tafaRU:
Steps to reproduce:
- create more than one variant
- change `fix_price` of one variant

Current behavior:
- `lst_price` of all variants are changed

Expected behavior:
- `lst_price` of only selected variant is changed

### DIFF
--- a/product_variant_sale_price/models/product_product.py
+++ b/product_variant_sale_price/models/product_product.py
@@ -23,7 +23,8 @@ class ProductTemplate(models.Model):
     def write(self, vals):
         res = super(ProductTemplate, self).write(vals)
         for template in self:
-            template._update_fix_price(vals)
+            if not self.env.context.get('skip_update_fix_price', False):
+                template._update_fix_price(vals)
         return res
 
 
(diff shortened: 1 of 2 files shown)


pedrobaeza:
@sergio-teruel please check

tafaRU:
FYI, this bug was introduced by https://github.com/OCA/product-variant/pull/37/commits/4a7511fbb2ddfbb1e1df388cf0e926210097cae4 via https://github.com/OCA/product-variant/pull/37

tafaRU:
@pedrobaeza, please set label as WIP.
My patch seeems do not work altough the bug is still there. :worried:

pedrobaeza:
OK

tafaRU:
I should have fixed :smiley: I confirm in a while.

tafaRU:
@pedrobaeza, you could remove the label.
Thank you.

pedrobaeza:
Done. Waiting for my colleague's opinion, who was the one that did the patch.

anajuaristi:
The single thing it must be done, is to make invisible the lst_price on template when this module is installed.
I will recomend to go back to the original version of the module and avoid all the code you have 
 included because it was working perfectly just avoiding to change the price from template.

sergio-teruel (approved):

PR #86, opened 2017-12-15
[IMP] product_variant_configurator: Add required flag to attribute lines

pedrobaeza:
With this, you can decide at product level which attributes are required to be filled. Default is true for keeping the same logic with previous version.

@Tecnativa


product_variant_configurator/models/product_product.py
```diff
@@ -83,20 +83,33 @@ def _check_configuration_validity(self):
...
             return
+        attribute_line_model = self.env['product.attribute.line']
         for product in self:
-            if bool(product.product_tmpl_id.attribute_line_ids.mapped(
-                    'attribute_id') -
-                    product.attribute_value_ids.mapped('attribute_id')):
+            errors = []
+            attribute_lines = product.product_tmpl_id.attribute_line_ids
+            for value in product.attribute_value_ids:
+                line = attribute_line_model.search([
```
lmignon: @pedrobaeza Your implementation is not scalable and will introduce performance issue on products import.
You should avoid to call `search` into a double loop!

lmignon (changes requested):

product_variant_configurator/models/product_product.py
pedrobaeza: Any suggestion for doing this?

lmignon: @pedrobaeza 
Why not 
```python
if product.product_tmpl_id.attribute_line_ids.mapped('attribute_id').filtered(lambda a: a.required) -  product.attribute_value_ids.mapped('attribute_id')):
    raise ...
```

pedrobaeza: Yeah, it works with a slight variation (first `filtered` and then `mapped`). Thanks for the suggestion.

lmignon (approved):
@pedrobaeza Thank you for the change. (Code review only)

pedrobaeza:
@lmignon, are you interested on having this on 9.0? 

I also patched the 10.0 version of purchase_variant_configurator for allowing to add products after confirmation of the purchase order. See https://github.com/OCA/product-variant/commit/aa6957eae988933bcdfc3a1652749beda2edd24d#diff-8ba0538db21e8a2e295501eff9089f80R17

chienandalu (approved):
Looks good :+1: (although couldn't make functional test on runbot)

luismontalba (approved):
Tested on runbot

PR #88, opened 2018-02-15
[9.0][FIX]product_variant_configurator: fix create_variant_ids

zakiuu:
In standard, when attributes are added to an already created product
template, the variant without attributes is deleted or deactivated (been
used somewhere else) automatically.
In product variant configurator module we loosing that behavior when the
creation of the variants are delayed (i.e case where no auto variants are created).
This PR aims to restore that behavior by deactivating the variant with
no attributes.

### DIFF
--- a/product_variant_configurator/models/product_template.py
+++ b/product_variant_configurator/models/product_template.py
@@ -28,11 +28,13 @@ class ProductTemplate(models.Model):
             # avoid the warning when opening a new form in create
             # mode (ie when the onchange triggers when Odoo sets
             # the default value)
+            msg = _("Changing this parameter may cause automatic variants "
+                    "creation (in case variants are created automatically),"
+                    "or automatic variants deactivation (in case variants "
+                    "are created manually).")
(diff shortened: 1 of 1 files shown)


andreparames:
LGTM :+1:

andreparames (approved):

pedrobaeza (changes requested):
If you change this, then the warning on the onchange of the field should be updated for reflecting this.

zakiuu:
@pedrobaeza thanks for your review, I've changed. Hope it looks clear now

pedrobaeza:
I'm afraid if doesn't serve, as you have change the message inside the `if` clause. Please put 2 different messages depending on the value.

pedrobaeza:
Are you going to continue this?

ThomasBinsfeld:
@pedrobaeza I would like to finish this one. Could you explain me what must be done?

pedrobaeza (approved):

Issue #92, opened 2018-03-04
[10.0] Failed to write on multi record when product_default_variant_code is installed


sebastienbeau:
@bealdav @pierrickbrun @daniel-CA  @chienandalu  @gdgellatly  @pedrobaeza 

I have detected an issue with the module product_default_variant_code
Indeed the write on multi-record is broken. 

If you take a look here : https://github.com/OCA/product-variant/blob/10.0/product_variant_default_code/models/product.py#L129 the code check "self.attribute_line_ids" or "self" can be a list of record.

I do not have the time to propose I fix right now. But maybe next week

gdgellatly:
@sebastienbeau I don't really see a case where this field should be written to multiple records. Effectively updating multiple product templates to have the same mask?

Anyway if it helps this is how I handled the case. https://github.com/odoonz/odoonz-addons/blob/11.0/product_code_builder/models/product_template.py#L57

pedrobaeza:
Any news about this?

bealdav:
Same feedback with another use case when I delete products

```
File "/workspace/parts/product-variant/product_variant_default_code/models/product.py", line 129, in write
    if self.attribute_line_ids:
  File "/workspace/parts/odoo/odoo/fields.py", line 904, in __get__
    record.ensure_one()
  File "/workspace/parts/odoo/odoo/models.py", line 4823, in ensure_one
    raise ValueError("Expected singleton: %s" % self)
ValueError: Expected singleton: product.template(26, 27, 28, 29)
```

Daniel-CA:
@bealdav How did you get this error? Just deleting the selected product_templates?

bealdav:
yes

bealdav:
I will fix it

pedrobaeza:
Closing as the PR fixing it was merged.

PR #94, opened 2018-03-16
[10.0] product_variant_default_code : empty separator + codes ordering

PierrickBrun:
This PR allows to use an empty separator by using the "None" string in the system parameter. It is impossible to set a system parameter to an empty string so if we need to use an empty separator we can use "None" as a string.

Moreover it forces the sorting order on the attribute sequence. It is possible to easily overwrite it because the _get_attr_val_k function is one line long.

### DIFF
--- a/product_variant_default_code/README.rst
+++ b/product_variant_default_code/README.rst
@@ -80,6 +80,11 @@ Unset `manual code` and the reference code will be unlocked again.
    :alt: Try me on Runbot
    :target: https://runbot.odoo-community.org/runbot/137/10.0
 
+Change the reference separator
+------------------------------
+
+You can change the separator using the system parameter named "default_reference_separator".
+If you do not want to have a separator (an empty one), you can the the parameter to "None".
 
 Known issues / Roadmap
 ======================
--- a/product_variant_default_code/README.rst
+++ b/product_variant_default_code/README.rst
@@ -76,15 +76,15 @@ variant internal reference will no longer be changed by this module.
 
 Unset `manual code` and the reference code will be unlocked again.
 
-.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
-   :alt: Try me on Runbot
-   :target: https://runbot.odoo-community.org/runbot/137/10.0
-
 Change the reference separator
 ------------------------------
 
 You can change the separator using the system parameter named "default_reference_separator".
-If you do not want to have a separator (an empty one), you can the the parameter to "None".
+If you do not want to have a separator (an empty one), you can set the parameter to "None".
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/137/10.0
 
 Known issues / Roadmap
 ======================
--- a/product_variant_default_code/models/product.py
+++ b/product_variant_default_code/models/product.py
@@ -103,7 +103,7 @@ class ProductTemplate(models.Model):
 
     def _get_default_mask(self):
         if not self.attribute_line_ids:
-            return None
+            return False
         attribute_names = []
         default_reference_separator = self.env[
             'ir.config_parameter'].get_param('default_reference_separator')
@@ -191,7 +191,6 @@ class ProductAttribute(models.Model):
 
     def write(self, vals):
         result = super(ProductAttribute, self).write(vals)
-
         if 'code' not in vals:
             return result
         # Rewrite reference on all product variants affected
--- a/product_variant_default_code/models/product.py
+++ b/product_variant_default_code/models/product.py
@@ -98,11 +98,17 @@ class ProductTemplate(models.Model):
              '\nNote: make sure characters "[,]" do not appear in your '
              'attribute name')
 
+    def _get_attr_val_k(self, attr_value_id):
+        return attr_value_id.attribute_id.sequence
+
     def _get_default_mask(self):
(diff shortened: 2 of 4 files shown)


bealdav (approved):

gurneyalex (changes requested):
Tests are broken

PierrickBrun:
@gurneyalex Tests are fixed now

PierrickBrun:
This issue seems fixed https://github.com/OCA/product-variant/issues/92 as test n°13 is passing

sebastienbeau:
Review + test

Now the following issue are fixed
- mask was not updated when changing the code_prefix
- mask are correctly updated when changing the attribute
- write on multi-product-template was broken

sebastienbeau (approved):

sebastienbeau:
@pedrobaeza this change are ok for you ? We will need it for the next PR : https://github.com/OCA/product-variant/pull/95

product_variant_default_code/README.rst
```diff
@@ -80,6 +80,11 @@ Unset `manual code` and the reference code will be unlocked again.
    :alt: Try me on Runbot
    :target: https://runbot.odoo-community.org/runbot/137/10.0
 
+Change the reference separator
```
pedrobaeza: Please put this before image

product_variant_default_code/README.rst
```diff
@@ -80,6 +80,11 @@ Unset `manual code` and the reference code will be unlocked again.
...
+------------------------------
+
+You can change the separator using the system parameter named "default_reference_separator".
+If you do not want to have a separator (an empty one), you can the the parameter to "None".
```
pedrobaeza: s/the the/set the

product_variant_default_code/models/product.py
```diff
@@ -98,44 +98,54 @@ class ProductTemplate(models.Model):
...
+            '.group_product_default_code')
+
     @api.model
     def create(self, vals):
         product = self.new(vals)
-        if (not vals.get('reference_mask') and product.attribute_line_ids or
-                not self.user_has_groups(
-                    'product_variant_default_code.group_product_default_code'
-                )):
+        if self._is_automatic_mask() or not vals.get('reference_mask'):
```
pedrobaeza, line 127: Why do you remove the `product.attribute_line_ids` check?

product_variant_default_code/models/product.py
```diff
@@ -98,44 +98,54 @@ class ProductTemplate(models.Model):
...
-                )):
-            if self.attribute_line_ids:
-                vals['reference_mask'] = self._get_default_mask()
         result = super(ProductTemplate, self).write(vals)
+        if (self._is_automatic_mask() and self._mask_needs_update(vals)) \
+                or ('reference_mask' in vals and not vals['reference_mask']):
+            for record in self:
+                # we write the new mask and the additional write
+                # will recompute the variant code
+                record.reference_mask = record._get_default_mask()
```
pedrobaeza, line 143: I don't like too much that there are several nested writes. I prefer that you make this operation before, copying vals and adding the changed value, and call `super` record by record.

product_variant_default_code/models/product.py
```diff
@@ -180,9 +190,10 @@ class ProductAttribute(models.Model):
         ('number_uniq', 'unique(name)', _('Attribute Name must be unique!'))]
 
     def write(self, vals):
-        if 'code' not in vals:
-            return super(ProductAttribute, self).write(vals)
         result = super(ProductAttribute, self).write(vals)
+
```
pedrobaeza: Don't add empty blank lines in the middle of a method

pedrobaeza (changes requested):

product_variant_default_code/models/product.py
PierrickBrun: The check is done in the _get_default_mask() function, so that we do not have to write it twice in the create and in the write

product_variant_default_code/models/product.py
```diff
@@ -98,44 +98,54 @@ class ProductTemplate(models.Model):
              '\nNote: make sure characters "[,]" do not appear in your '
              'attribute name')
 
+    def _get_attr_val_k(self, attr_value_id):
+        return attr_value_id.attribute_id.sequence
+
     def _get_default_mask(self):
+        if not self.attribute_line_ids:
+            return None
```
pedrobaeza: return False

product_variant_default_code/models/product.py
pedrobaeza: OK, right.

product_variant_default_code/models/product.py
sebastienbeau: Hi Pedro, I do not like a lot also the nested write, but in that case I think it's the simplest solution.
As for computing the "_get_default_mask" we need to have the attribute_line_ids wroten so the call to super must be done before getting the mask.

In the previous version if you modify the field "attribute_line_ids" the compute of the mask was done before the call to super and was generating a wrong mask with the previous value of the "attribute_line_ids".
What do you think?

pedrobaeza: OK, thanks for the clarification. I didn't see that. Then the only solution is to nest the write, yes.

sebastienbeau: Can you approve the review? thanks

pedrobaeza (approved):

pedrobaeza:
Merging as @gurneyalex's comment about the tests is fixed.

sebastienbeau:
Thanks Pedro !

PR #102, opened 2018-09-03
11.0 mig product variant available in pos

SimoRubi:
Standard migration


crgb001:
Question: Can the attribute order defined in the product template be changed? 
Currently the first attribute will show as columns and the second as rows. Due to the initial configuration of the products in our implementation, the attribute with less values and different from each product(COLOR), shows in the columns, while the attribute with more values and more stable among the products (SIZE) shows as rows. Although this doesnt affect the operations it is a little ackward.
The problem is that in order to change this I would have to modify each product template (>800), deleting the attributes and introducing them again in the order i need.
While reviewing the code where the  "2D matrix for x2many fields" module is implemented (sale_manage_variant.py) I saw that the code is defined as follows:

        if template and len(template.attribute_line_ids) >= 2:
            line_x = template.attribute_line_ids[0]
            line_y = template.attribute_line_ids[1]

¿If I change (in my local files) to ...

        if template and len(template.attribute_line_ids) >= 2:
            line_x = template.attribute_line_ids[1]
            line_y = template.attribute_line_ids[0]

will I revert the order of the matrix? do i need to make more changes? will this affect the way the code works?

Maybe a configuration file can be added in future implementations where the user can specify which attribute should be columns and which one rows.

Thanks everyone for the great effort. This module is one of the best tools when using variants.

SimoRubi:
Hi @crgb001, this thread is dedicated to the migration to v11 of module product_variant_available_in_pos.

Your question is actually about a functionality that is introduced by module [sale_order_variant_mgmt](https://github.com/OCA/product-variant/tree/11.0/sale_order_variant_mgmt) or, more precisely, by module [web_widget_x2many_2d_matrix](https://github.com/OCA/web/tree/11.0/web_widget_x2many_2d_matrix) that unfortunately I'm not very familiar with.

For requests/issues about those modules please post an issue using the link *GitHub Issues* that is placed in the module's page.

tafaRU (approved):

eLBati (approved):

SimoRubi:
@OCA/product-maintainers could you please check? Thanks

product_variant_available_in_pos/__manifest__.py
```diff
@@ -1,15 +1,15 @@
...
 {
     'name': 'Product Variant Available In Pos',
     'summary': 'Available In Pos in product level',
-    'version': '8.0.1.0.0',
+    'version': '11.0.1.0.0',
     'category': 'Point Of Sale',
     'license': 'AGPL-3',
     'author': 'Agile Business Group, Odoo Community Association (OCA)',
-    'website': 'https://www.agilebg.com/',
+    'website': 'https://github.com/OCA/product-variant/tree/'
```
pedrobaeza, line 11: Please use repository URL, not full one

product_variant_available_in_pos/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,5 @@
+This module allows to have 'Available in the Point of Sale' flag at
+product level instead of product template.
+
+Please note this module is incompatible with `pos_product_template
```
pedrobaeza: Put this in ROADMAP

pedrobaeza (changes requested):

product_variant_available_in_pos/__manifest__.py
SimoRubi: I actually prefer the full link, so when someone clicks on it, they directly see the module's (brand new) README. 
Is there any side effect in using the full link?

product_variant_available_in_pos/readme/DESCRIPTION.rst
SimoRubi: Thanks for the review!
That's seems right, done in https://github.com/OCA/product-variant/pull/102/commits/2528411d0a4974ecf7167b363cf40d9c46de302c

product_variant_available_in_pos/__manifest__.py
pedrobaeza: It forces you to change it each time you migrate the module. The README is displayed at the same time of the manifest, so I don't see this very useful.

SimoRubi:
Fixed up commits

tafaRU:
@pedrobaeza, could you please update your review?

product_variant_available_in_pos/models/product_product.py
```diff
@@ -0,0 +1,24 @@
...
+    @api.model
+    def create(self, vals):
+        product = super(ProductProduct, self).create(vals)
+        template_vals = {}
+        if 'available_in_pos' not in vals:
+            template_vals[
+                'available_in_pos'] = product.product_tmpl_id.available_in_pos
+        if template_vals:
+            product.write(template_vals)
+        return product
```
espo-tony, line 20: This surely works, but I find it a bit confusing. May I suggest to rewrite it as:

```
    @api.model
    def create(self, vals):
        product = super(ProductProduct, self).create(vals)
        if 'available_in_pos' not in vals:
            product.available_in_pos = product.product_tmpl_id.available_in_pos
        return product
```

product_variant_available_in_pos/models/product_template.py
```diff
@@ -0,0 +1,29 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, models
+
+
+class ProductTemplate(models.Model):
+
+    _inherit = 'product.template'
+
+    def _update_available_in_pos(self, vals):
```
espo-tony, line 12: Please, add an `ensure_one()` here, as the method is meant to be used only on a single record at time.

product_variant_available_in_pos/models/product_template.py
```diff
@@ -0,0 +1,29 @@
...
+    @api.multi
+    def write(self, vals):
+        res = super(ProductTemplate, self).write(vals)
+        if 'available_in_pos' in vals:
+            for product in self:
+                product._update_available_in_pos(vals)
+        return res
+
+    @api.model
+    def create(self, vals):
```
espo-tony: Are you sure this is needed? I think `product.product` `create` method already takes care of it.

espo-tony (changes requested):

product_variant_available_in_pos/models/product_product.py
SimoRubi: Thanks for your review @espo-tony , I did not any refactoring because this is only a migration.
Anyway I added a new commit to add what you asked for: https://github.com/OCA/product-variant/pull/102/commits/83cc3ddb450f087e313212bce8ae0c4c57928967

espo-tony (approved):
LGTM :+1:

product_variant_available_in_pos/__manifest__.py
pedrobaeza: This link is not working anyway.

SimoRubi: @pedrobaeza the link should work when the PR is merged, it is the same as https://github.com/OCA/product-variant/tree/8.0/product_variant_available_in_pos
but just with 11.0 instead of 8.0

SimoRubi:
@pedrobaeza if https://github.com/OCA/product-variant/pull/102#discussion_r216680457 is not blocking, could you please change your review? thanks

product_variant_available_in_pos/__manifest__.py
pedrobaeza: OK, I don't get why do you prefer this long form, but I don't block more.

pedrobaeza (approved):

PR #109, opened 2018-10-11
[11.0] [MIG] purchase_order_variant_mgmt

javierjcf:


pedrobaeza:
Please squash together adjacent "OCA Transbot.." commits using https://github.com/OCA/maintainer-tools/wiki/Merge-commits-in-pull-requests

javierjcf:
It's squashed now.

pedrobaeza:
There is another migration in progress in https://github.com/OCA/purchase-workflow/pull/594, but this is the right repository.

purchase_order_variant_mgmt/wizard/purchase_manage_variant_view.xml
```diff
@@ -18,10 +18,15 @@
                        field_x_axis="value_x"
                        field_y_axis="value_y"
                        field_value = "product_uom_qty"
```
grindtildeath: Please fix the space here.

purchase_order_variant_mgmt/wizard/purchase_manage_variant.py
```diff
@@ -91,16 +90,13 @@ def button_transfer_to_order(self):
                 order_line._onchange_quantity()
                 order_line_vals = order_line._convert_to_write(
                     order_line._cache)
-                purchase_order.order_line.create(order_line_vals)
+                purchase_order.order_line.browse().create(order_line_vals)
```
grindtildeath: Is the `browse` here really needed ? IMO would look cleaner to use `self.env['purchase.order.line']`

purchase_order_variant_mgmt/tests/test_purchase_order_variant_mgmt.py
```diff
@@ -92,10 +99,9 @@ def test_modify_variants(self):
             len(wizard.variant_line_ids.filtered('product_uom_qty')), 2,
             "There should be two fields with any quantity in the wizard."
         )
-        wizard.variant_line_ids.filtered(
-            lambda x: x.product_id == product1).product_uom_qty = 0
-        wizard.variant_line_ids.filtered(
-            lambda x: x.product_id == product2).product_uom_qty = 10
+
+        wizard.variant_line_ids[0].product_uom_qty = 0
+        wizard.variant_line_ids[1].product_uom_qty = 10
```
grindtildeath, line 106: Was this change really needed ? IMO looks safer to assert a quantity according to its product than its order in a recordset.

grindtildeath (commented):
Thanks for the port, I tested and works well. Just a few minor things here and can you please split the readme into parts ?

Then, this is not really to do here in the migration, but since it appears here I state it : it would be great to change the message "Sorry no matrix data to display." to something more specific to the context like "Select a product first" here. However it seems there's no hook to define a specific message in the matrix widget module, but having the possibility to set the message in the XML field definition would be IMO a very nice improvement to the widget. cc @simahawk @arkostyuk

purchase_order_variant_mgmt/wizard/purchase_manage_variant_view.xml
javierjcf: fixed

purchase_order_variant_mgmt/wizard/purchase_manage_variant.py
javierjcf: I did the change because in the sale_order_variant_mgmt (already migrated) appears that way. I don't know the reason and i copy it. It is with the self.env['purchase.order.line'] now.

purchase_order_variant_mgmt/wizard/purchase_manage_variant_view.xml
javierjcf: Readme folder was added.

purchase_order_variant_mgmt/tests/test_purchase_order_variant_mgmt.py
javierjcf: I revert the cahnge.

purchase_order_variant_mgmt/wizard/purchase_manage_variant.py
pedrobaeza: It's needed, yes, because you will have a singleton error if not when having existing lines. It's a curious side effect here in v11.

purchase_order_variant_mgmt/tests/test_purchase_order_variant_mgmt.py
javierjcf: I did this change again. because now, the wizard line has not field product_id, so the filtered was raising an error.

pedrobaeza: I have written about the same, but you are right. The product in current 2D matrix widget can be stored, so this is the less worst solution.

pedrobaeza: @simahawk this is a flaw in current matrix widget: you can't store a full record in the matrix. Only the fields in row, column and value are stored.

purchase_order_variant_mgmt/wizard/purchase_manage_variant.py
```diff
@@ -91,16 +90,13 @@ def button_transfer_to_order(self):
...
         lines2unlink.unlink()
 
 
 class PurchaseManageVariantLine(models.TransientModel):
     _name = 'purchase.manage.variant.line'
 
-    product_id = fields.Many2one(
-        comodel_name='product.product', string="Variant", readonly=True)
-    disabled = fields.Boolean()
```
javierjcf: The product_id field was removed, so in the test file we can't do:
`  wizard.variant_line_ids.filtered(lambda x: x.product_id == product1).product_uom_qty = 0 `

pedrobaeza: Yeah, I saw it later on your other comment. Let's continue the conversation there.

javierjcf:
> This was done on purpose for not being slave of the insertion order. Why do you remove it?

Because the product_id field was removed from _purchase.manage.variant.line_, so in the test file we can't do:
wizard.variant_line_ids.filtered(lambda x: x.product_id == product1).product_uom_qty = 0

purchase_order_variant_mgmt/tests/test_purchase_order_variant_mgmt.py
simahawk: wild guess: can't we bypass this by making these fields invisible?
Otherwise: can you open an issue on OCA/web?

pedrobaeza: I'm afraid it doesn't serve. We were discussing in this PR: https://github.com/OCA/web/pull/1022, although it was more an internal chat between my colleague @Yajo and I.

purchase_order_variant_mgmt/wizard/purchase_manage_variant.py
javierjcf: Ahhh, OK i underestand the empty browse now, but with self.env['purchase.order.line'].create() we avoid any singleton error because we don't need to acces to purchase_order.order_line instance to create. I think that using self.env[] is the normal way to create a record.

pedrobaeza: Yeah, that's another option indeed. Here we make the browse for not having to enumerate the exact model, as we have it in the field, but if you see it more readable, please change it.

purchase_order_variant_mgmt/tests/test_purchase_order_variant_mgmt.py
simahawk: FTR reported here https://github.com/OCA/web/issues/1084

purchase_order_variant_mgmt/__manifest__.py
```diff
@@ -0,0 +1,24 @@
...
+{
+    'name': 'Handle easily multiple variants on Purchase Orders',
+    'summary': 'Handle the addition/removal of multiple variants from '
+               'product template into the purchase order',
+    'version': '11.0.1.0.0',
+    'author': 'Tecnativa,'
+              'Odoo Community Association (OCA)',
+    'category': 'Purchases',
+    'license': 'AGPL-3',
+    'website': 'https://www.tecnativa.com',
```
simahawk, line 13: https://github.com/OCA/product-variant

purchase_order_variant_mgmt/__manifest__.py
```diff
@@ -0,0 +1,24 @@
...
+    'depends': [
+        'purchase',
+        'web_widget_x2many_2d_matrix',
+    ],
+    'demo': [],
```
simahawk: remove demo

purchase_order_variant_mgmt/README.rst
```diff
@@ -0,0 +1,83 @@
+
```
simahawk: for the next reviewer: this will be replaced by README generator ;)

purchase_order_variant_mgmt/tests/test_purchase_order_variant_mgmt.py
```diff
@@ -0,0 +1,111 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo.tests import common
+
+
+class TestPurchaseOrderVariantMgmt(common.SavepointCase):
+    @classmethod
+    def setUpClass(cls):
+        super(TestPurchaseOrderVariantMgmt, cls).setUpClass()
+        cls.partner = cls.env['res.partner'].create({'name': 'Test partner'})
```
simahawk, line 11: disable tracking on the whole suite:
```py
cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))
```

purchase_order_variant_mgmt/wizard/purchase_manage_variant.py
```diff
@@ -0,0 +1,104 @@
+# Copyright 2016 Pedro M. Baeza <pedro.baeza@tecnativa.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+import openerp.addons.decimal_precision as dp
```
simahawk: odoo

purchase_order_variant_mgmt/wizard/purchase_manage_variant.py
```diff
@@ -0,0 +1,104 @@
...
+from odoo import api, fields, models
+
+
+class PurchaseManageVariant(models.TransientModel):
+    _name = 'purchase.manage.variant'
+
+    product_tmpl_id = fields.Many2one(
+        comodel_name='product.template', string="Template", required=True)
```
simahawk: would be nice to have params on new lines

purchase_order_variant_mgmt/wizard/purchase_manage_variant.py
```diff
@@ -0,0 +1,104 @@
...
+                    'product_uom_qty': line.product_uom_qty,
+                    'order_id': purchase_order.id,
+                })
+                order_line.onchange_product_id()
+                # This should be done later for handling supplier quantities
+                order_line.product_qty = line.product_uom_qty
+                order_line._onchange_quantity()
+                order_line_vals = order_line._convert_to_write(
+                    order_line._cache)
+                self.env['purchase.order.line'].create(order_line_vals)
```
simahawk: wouldn't be better to collect line values as `(0, 0, vals)` and write on the order once?

simahawk (changes requested):

purchase_order_variant_mgmt/README.rst
pedrobaeza: I prefer to pre-generate the README on the migration PR, as you reduce diff size and avoid these confusions. @javierjcf maybe you can try to install maintainer-tools (with python3) and run oca-gen-addon-readme for having it.

javierjcf:
I don't understand why now ist's raising a flake8 error in the travis integretion, in the product_variant_sale_price module, which is not changed in this PR.

this is the error:
`product_variant_sale_price/hooks.py:6: [E8103(sql-injection), set_sale_price_on_variant] SQL injection risk. Use parameters if you can. - More info https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#no-sql-injection`

pedrobaeza:
That warning was before, but it's not the one that provokes the red state. This is the error:
```
purchase_order_variant_mgmt/wizard/purchase_manage_variant.py:16:53: W291 trailing whitespace
```

javierjcf:
> That warning was before, but it's not the one that provokes the red state. This is the error:
> 
> ```
> purchase_order_variant_mgmt/wizard/purchase_manage_variant.py:16:53: W291 trailing whitespace
> ```

Thanks @pedrobaeza. It's fixed now.

GSLabIt:
is this ready to merge?

pedrobaeza:
There are no approvals on the PR, so it should be reviewed and approved before being merged. @GSLabIt you can be one of them.

GSLabIt:
@pedrobaeza will have a look in a few, thx

GSLabIt (approved):
Functional tests LGTM

pedrobaeza:
@javierjcf please squash your migration commits together for merging this.

grindtildeath (commented):
Tested and looks ok, however, the widget is limited for products having more than 2 variants, shouldn't we state this clearly in readme/ROADMAP.rst and maybe restrict the selection of product template to products having max 2 variants ?

purchase_order_variant_mgmt/__manifest__.py
grindtildeath: Still to attend

pedrobaeza:
Superseeded by #152

pedrobaeza:
Not really superseeded, so reopening.

Roodin:
LGTM

pedrobaeza:
Let's merge it as an old one and already migrated to upper versions: /ocabot merge

PR #113, opened 2018-10-25
[FIX] sale_variant_configurator: typo, field does not exist

cubells:
Typo

cc @Tecnativa


pedrobaeza (approved):
Fast-tracking this obvious patch

PR #114, opened 2018-10-30
[FIX] 10.0 product_variant_default_code multi-company support

PierrickBrun:
Before this fix, if you update a product attribute value code, only the products (default_code) of you company will be updated accordingly.

This fix corrects this behaviour.

For reference : https://github.com/OCA/product-attribute/pull/380


bealdav (approved):

pedrobaeza (approved):

pedrobaeza:
Please forward-port it to v11

PR #115, opened 2018-11-16
[MIG] sale_stock_variant_configurator: Migration to 11.0

cubells:
* Update read files
* simple migration

cc @Tecnativa


pedrobaeza:
This one shouldn't be needed anymore.

PR #116, opened 2018-11-26
[IMP] sale_order_variant_mgmt: Default values are not applied with new

pedrobaeza:
This is a flaw/design decision on `new` method, so we have to call it explicitly for getting same result as real records.

@Tecnativa


bodedra (approved):
:heavy_check_mark:

yajo (approved):

PR #122, opened 2018-12-05
[FIX] *_variant_configurator: attribute lines are not created

cubells:
Propagate context in order to product attribute lines are created.

cc @Tecnativa

### DIFF
--- a/purchase_variant_configurator/__manifest__.py
+++ b/purchase_variant_configurator/__manifest__.py
@@ -7,7 +7,7 @@
 {
     "name": "Purchase - Product variants",
     "summary": "Product variants in purchase management",
-    'version': '10.0.1.1.0',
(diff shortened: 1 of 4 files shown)


pedrobaeza:
@SalahAdDin check this one for solving the creation on the fly of the attribute.

SalahAdDin:
@pedrobaeza Checked, i can see it works like a charm!
deepin-screen-recorder_select area_20181207161210

Tested in Odoo 11.0..

pedrobaeza:
OK, we are fixing this in v10, and then cherry-picking it in v11.

pedrobaeza:
Cherry-picked for v11 in a457c51a683ef85e4f8f644c140ec0dad74ec93b

Issue #120, opened 2018-12-04
[11.0][sale_variant_configurator] Remose the user interface


SalahAdDin:
Hi,

I'm seeing the current interface and i think it could be better:
![deepin-screen-recorder_select area_20181204142559](https://user-images.githubusercontent.com/5034215/49440948-23c56080-f7d6-11e8-9bbc-daf72a3879bd.gif)

It is not needed to handle all line product in the new view, it just requires having a select box to select the product and the product's attribute fields. Also, creating a new attribute value does not require a new form since the attribute must to be the current attribute and the other fields(sequence and attribute sequence) can be filled automatically.
 
What do you think about that?

pedrobaeza:
Why do you put again and again the same video?

It's done this way on purpose for a lot of reasons:

* Guide the introduction of each attribute isolatedly. Other way, it would be very difficult to recognize each one.
* You don't have to deal with multiple selection of several values for the same attribute
* Not all attribute must be filled and this way it's very easy to be recognized.

SalahAdDin:
But, at least the attribute value's form is not required, because, if you are creating a new value for a determinate attribute, you should not choose it again, the system should fill it automatically.

pedrobaeza:
I don't get what do you mean with that.

SalahAdDin:
Actually, I think the form for purchases is more aesthetic than the sale order form:
![esteestamejor](https://user-images.githubusercontent.com/5034215/49444962-b9b2b880-f7e1-11e8-9462-afcca8fb01de.png)

Como me puedo explicar, cuando creamos un nuevo valor para el atributo, este nos envía a un formulario nuevo donde creamos ese nuevo valor, pero aquí debemos elegir el atributo al cuál le vamos a crear un valor, pero por qué? Este debería ser elegido automáticamente, desde que el formulario se lanza desde un atributo existente; es innecesario volver a elegir el atributo.

pedrobaeza:
The idea is to not go to that popup form, but directly choose "Create" option, not "Create and edit...".

SalahAdDin:
I did it, but it triggers the _Create and Edit_ popup.

pedrobaeza:
Yeah, but this is current situation, not the final one.

SalahAdDin:
Hi, i have a possible suggestion:

[...]

PR #123, opened 2018-12-05
[ADD] mrp_variant_configurator: Migration to 11.0

SalahAdDin:
This is my first intent to add product configurator to the Manufacture module.

It still has a problem, i'm working to solve it.

TODO: 
- [ ] <del>Create products on change (form breaks when we edit the attributes' values).</del> General problem.
- [ ] <del>Proper description field (as **SO** description field).</del> General problem.

### DIFF
--- a/mrp_variant_configurator/__init__.py
+++ b/mrp_variant_configurator/__init__.py
@@ -1,0 +1,5 @@
+# Copyright 2017 Jose Luis Sandoval Alaguna <jose.alaguna@rotafilo.com.tr>
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+
+from . import models
+from .hooks import assign_product_template
--- a/mrp_variant_configurator/__manifest__.py
+++ b/mrp_variant_configurator/__manifest__.py
@@ -1,0 +1,23 @@
+# Copyright 2017 Jose Luis Sandoval Alaguna <jose.alaguna@rotafilo.com.tr>
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+{
+    'name': 'MRP - Product variants',
+    'version': '11.0.1.0.1',
+    'summary': 'Product variants in MRP management',
+    'description': '',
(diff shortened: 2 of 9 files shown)


SalahAdDin:
I found a first issue:

deepin-screen-recorder_select area_20181206122454
<del>While in **PO** it is not needed to create the product before at moment (don't check `create a product now!`), in **BoM** you must to do it before to save the _line_.
Why?</del>
It's a problem with configurator, we can find it in the default [module for purchases](https://github.com/OCA/product-variant/issues/120#issuecomment-444808756).

pedrobaeza:
Please preserve commit history following technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-11.0

SalahAdDin:
I did all steps, and I squashed all commits to create just one commit with the new module.

pedrobaeza:
But that's not the idea, as you are usurping the authorship of the rest. You can only squash commits from same author.

SalahAdDin:
But, i'm the author to this module, i mean, it does not exist in previous versions afaik.

pedrobaeza:
Ah, OK, then why do yo put `[MIG]` and try to add it in the migration issue? This is for migrations. In your case, it's an addition. Check https://www.odoo.com/documentation/11.0/reference/guidelines.html#tag-and-module-name for extra info.

pedrobaeza:
Why do you close it? I don't get your way of acting most of the time...

SalahAdDin:
I'm trying to change the tags to `ADD`.

Ok, i did: https://github.com/OCA/product-variant/pull/126

pedrobaeza:
You don't need to create a new PR for that. Just edit GitHub PR title. And if you want to amend commit title, just do it locally, and forced push the branch.

SalahAdDin:
Next time, i made a mistake: i deleted the branch and create a new one with the correct name. I'm sorry.

pedrobaeza:
OK, no problem, just passing by the learning process as everybody.

PR #133, opened 2019-01-24
[10.0] product_variant_default_code do not show default_code on template

PierrickBrun:
before the PR:
both reference_prefix and default_code are shown on the product_template (before it has variants)

This is confusing because only the reference_prefix should be used, also the default_code will be replaced if used.

after the PR:
only the reference prefix is shown on the template.
You can still view the default_code from the name_get()

It looks like it's what was tried with the attrs="{'invisible': True}".


bealdav (approved):

robinkeunen (approved):

PierrickBrun:
@pedrobaeza can you merge this please ? :smiley:

PR #134, opened 2019-01-30
[PERF] Make seller_ids and tmpl_seller_ids stored

PeterAlabaster:
When confirming complex sales orders, calls to `product.seller_ids`
in `name_get()` and `select_seller()` result in the ORM re-calculating this fields
values for many products due to running out of cache (i believe).

This result in ~6s compute time per line being wasted on re-running the lambdas.
In my case this compute function was being hit 800,000 times on a 8 line SO.
Making the field stored resolves this issue,
but obviously has some impact when initially installing the module.

There appears to be no detrimental effects to making this change,
once the module is installed.

### DIFF
--- a/product_variant_supplierinfo/__manifest__.py
+++ b/product_variant_supplierinfo/__manifest__.py
@@ -5,7 +5,7 @@
 {
     'name': 'Product supplier info per variant',
     'summary': 'Supplier info to product variant scope',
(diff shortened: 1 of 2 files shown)


PeterAlabaster:
After more testing, this does have detrimental effects, particularly on purchase order confirmation - sprouting an unusual error regarding UoM differences. Closing.

PR #137, opened 2019-02-26
[10.0][FIX] sale_variant_configurator: Does not return the inherited result

igallart:
By not returning the inherited result, some methods like https://github.com/OCA/partner-contact/blob/b2835680c05a9c1c3853c766604797b992b74ba8/partner_sale_risk/models/sale.py#L37 do not work and the user does not know what has happened.


pedrobaeza (approved):

pedrobaeza:
Fast-tracking as obvious

PR #138, opened 2019-03-11
[11.0][FIX] product_variant_configurator

AdriaGForgeFlow:
Fix to the following error that appears when searching for product template:
image

cc ~ @eficent


pedrobaeza:
Better to check for `limit` value instead, as `limit=None` means no limit, and you are hardcoding always a limit.

pedrobaeza:
I mean in line 100, to put ` if limit and len(res) >= limit:`

LoisRForgeFlow (approved):
LGTM :+1:

pedrobaeza (approved):

PR #142, opened 2019-04-29
[IMP] sale_order_variant_mgmt: Play all onchanges

pedrobaeza:
There are third-party modules that can add onchanges to the fields that this module is setting in sale.order.line, so this is the only way to ensure proper line creation.

This means the introduction of a new dependency in a stable branch, but it's worth IMO. What others think?

cc @Tecnativa

### DIFF
--- a/oca_dependencies.txt
+++ b/oca_dependencies.txt
@@ -1,1 +1,2 @@
+server-tools
 web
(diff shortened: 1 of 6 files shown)


yajo (approved):

rafaelbn:
Travis :red_circle: something is wrong in de OCA runbots I guess... :thinking:

pedrobaeza:
Closing as conflicting one.

PR #146, opened 2019-06-20
[12.0][MIG] product_variant_configurator

AdriaGForgeFlow:
Standard migration of module product_variant_configuration to v12.

cc ~ @Eficent


pedrobaeza:
Aren't you finally removing the feature of auto-create variants from this module?

AdriaGForgeFlow:
> Aren't you finally removing the feature of auto-create variants from this module?

I am not sure what to do, what are your thoughts?

I feel the core functionality is not straight forward. When a product has many attributes you would thing that variants would be created for those attributes that have _Always_ in the Create Variants field but, instead, if you have at least one attribute with _Only when a SO is created_ in the Create Variants field it will not create any variant at all. It does not feel right.

pedrobaeza:
@ageficent I'm on holidays till Wednesday, but we can have a meet for checking the design if you want then.

AdriaGForgeFlow:
> @ageficent I'm on holidays till Wednesday, but we can have a meet for checking the design if you want then.

Alright then :+1: Enjoy the holidays!

product_variant_configurator/models/product_configurator.py
```diff
@@ -0,0 +1,275 @@
...
+        comodel_name='product.configurator.attribute',
+        domain=lambda self: [("owner_model", "=", self._name)],
+        inverse_name='owner_id', string='Product attributes', copy=True)
+    price_extra = fields.Float(
+        compute='_compute_price_extra',
+        digits=dp.get_precision('Product Price'),
+        help="Price Extra: Extra price for the variant with the currently "
+             "selected attributes values on sale price. eg. 200 price extra, "
+             "1000 + 200 = 1200.")
+    product_id = fields.Many2one(
```
LoisRForgeFlow, line 31: This is generating an issue with inventory adjustments:
```
(...)
File "/home/lois/odoo/12.0/odoo/addons/stock/models/stock_inventory.py", line 280, in _get_inventory_lines_values
    GROUP BY product_id, location_id, lot_id, package_id, partner_id """ % domain, args)
(...)
psycopg2.ProgrammingError: column reference "product_id" is ambiguous
LINE 1: SELECT product_id, sum(quantity) as product_qty, location_id...
```
To reproduce just install stock and try to do a inventory adjustment

LoisRForgeFlow (changes requested):

product_variant_configurator/models/product_configurator.py
pedrobaeza: It's reported since a lot...

https://github.com/odoo/odoo/pull/34387

We will have to do fully qualified SQL expression in OCB. In master, it has been replaced by a proper ORM `read_group`.

elvise:
Hi, need any actions for merging...?

product_variant_configurator/__manifest__.py
```diff
@@ -0,0 +1,32 @@
+# Copyright 2015 Oihane Crucelaegui (AvanzOSC)
+# Copyright 2016 ACSONE SA/NV
+# Copyright 2017 David Vidal <david.vidal@tecnativa.com>
+# Copyright 2015-2017 Tecnativa - Pedro M. Baeza
```
tschanzt, line 4: Copyright 2019?

product_variant_configurator/__manifest__.py
```diff
@@ -0,0 +1,32 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    'name': 'Product Variant Configurator',
+    'summary': """
+        Provides an abstract model for product variant configuration.""",
```
tschanzt: i would either put it on single line or move down the end quotes

product_variant_configurator/models/product_template.py
```diff
@@ -0,0 +1,86 @@
...
+
+from odoo import _, api, fields, models
+from odoo.tools import config
+
+
+class ProductTemplate(models.Model):
+    _inherit = 'product.template'
+
+    no_create_variants = fields.Selection(
+        [('yes', "Don't create them automatically"),
```
tschanzt, line 14: maybe it would be better to invert this field or use no_create and create for the values.

tschanzt (approved):
looks good to me in general. two small points. and mabe the copyright dates in some files. but not really an issue for me.

AdriaGForgeFlow:
@pedrobaeza I was reviewing this PR and I'm yet not sure how to proceed... Would you keep the feature of variant creation introduced by this module or remove it and adapt to what is now in Odoo core?

pedrobaeza:
We should keep it, as it's not the same. Let's talk by chat about that.

elvise:
> @pedrobaeza I was reviewing this PR and I'm yet not sure how to proceed... Would you keep the feature of variant creation introduced by this module or remove it and adapt to what is now in Odoo core?

hi, i have tested it from runbot, i think it would be great if we can use it directly in the core configurator.

alvaro-domatix (approved):

bealdav (approved):

bealdav:
Nice job
Is it possible to push these last 2 commits to OCB ?
https://github.com/ForgeFlow/odoo/commits/12.0-temp-fix-variant_configurator

elvise:
Hi any good news about merge ?

pedroortega-punt:
I think that this module is absolutely necessary for a good operation with variants.

The need to decide in which product the automatic variants are created is unavoidable for the customer. keeping in mind that not a single independent variant of automatic creation can be created in the community. In the community there is no button on the product to create the variant without automation.

or is there some way to manually create a variant by assigning attributes without being automatic on the attribute?

When is their merger planned?

pedroortega-punt:
I have checked the module and I think that the module does not solve the problems that can be had when working with variants.

As the attributes are designed when creating the variants, I think it is not appropriate, since when creating attributes you will never know if in the template we are creating we will need to create all the variants or only when they are sold . This attribute for example, Color can be used in thousands of templates, and if we indicate the option to create "always" there will be templates that due to their great combination of attributes we do not want to create all the variants and others, so it would be very convenient to decide In which template will all the combinations of variants be created in order to indicate different costs.

As the attribute settings are created now you cannot decide which ones are going to be created and in which templates. Now only the order option fits, never or always in all the templates, so it can cause a collapse of the bbdd if we create templates with many attributes and many values.

**The module should be as before, have the option to decide**, which with the module removing the option to "Create them automatically" in the template and replacing it with "Use Oddo's default variant management" we are in the same option as with what we now provide the attributes.

Imagine that in this example we indicated in the attributes that the variants are always created.

imagen


However, there are other products that we need to create because we buy them at different cost and supplier, when the combinations are not thousands and you share the same attributes.

ThomasBinsfeld:
What's the status of this?

pedrobaeza:
I'm not able to review it now, so if it's OK for you, go ahead.

ThomasBinsfeld (approved):

jarroyomorales (approved):

elvise:
Hi guys, what's the status of this ?

elvise:
Hi, is there anyone available for review...?

HviorForgeFlow (approved):
Functional + Code Review 
LGTM :+1:

rousseldenis:
Hi @AdriaGForgeFlow is this ready ?

AdriaGForgeFlow:
> Hi @AdriaGForgeFlow is this ready ?

Yes, it is from my side :+1:

Issue #148, opened 2019-06-26
[11.0] sale_variant_configurator: pricelist error


cubells:
If pricelist discount_policy is equal to 'without_discount', and you search for a template in sale order line, then an error is raised because this line:

https://github.com/OCA/product-variant/blob/11.0/sale_variant_configurator/models/sale_order.py#L143

generate a list of tuples with a product.product recordset null, as product_id field is empty at this moment:

https://github.com/OCA/product-variant/blob/11.0/product_variant_configurator/models/pricelist.py#L21

To reproduce the error only have to change the discount_policy of the pricelist of the sale order and search for a template with variants in the sale order line.

The log I've got:

> 
> 
> 2019-06-26 16:49:49,911 25905 ERROR database odoo.sql_db: bad query: b"SELECT item.id FROM product_pricelist_item AS item LEFT JOIN product_category AS categ ON item.categ_id = categ.id WHERE (item.product_tmpl_id IS NULL OR item.product_tmpl_id = any(ARRAY[false]))AND (item.product_id IS NULL OR item.product_id = any(ARRAY[false]))AND (item.categ_id IS NULL OR item.categ_id = any('{}')) AND (item.pricelist_id = 1) AND (item.date_start IS NULL OR item.date_start<='2019-06-26 16:28:12') AND (item.date_end IS NULL OR item.date_end>='2019-06-26 16:28:12')ORDER BY item.applied_on, item.min_quantity desc, categ.parent_left desc"
> ERROR: operator does not exist: integer = boolean
> LINE 1: ...m.product_tmpl_id IS NULL OR item.product_tmpl_id = any(ARRA...
>                                                              ^
> HINT:  No operator matches the given name and argument type(s). You might need to add explicit type casts.
> 
> 2019-06-26 16:49:49,912 25905 ERROR database odoo.http: Exception during JSON request handling.
> Traceback (most recent call last):
>   File "/home/cubells/Dades/git/odoo11/src/core/odoo/http.py", line 653, in _handle_exception
>     return super(JsonRequest, self)._handle_exception(exception)
>   File "/home/cubells/Dades/git/odoo11/src/core/odoo/http.py", line 312, in _handle_exception
>     raise pycompat.reraise(type(exception), exception, sys.exc_info()[2])
>   File "/home/cubells/Dades/git/odoo11/src/core/odoo/tools/pycompat.py", line 87, in reraise
>     raise value
[...]

PR #149, opened 2019-06-27
[FIX] sale_variant_configurator: error depending of discount policy i…

cubells:
…n pricelist

Solves https://github.com/OCA/product-variant/issues/148


pedrobaeza:
Please split the README by fragments change in a separate commit.

cubells:
@pedrobaeza done.

pedrobaeza:
The main problem of this is that you are using template as a product, and their IDs might not be the same, causing problems. You should neutralize the part that uses products and provide an alternative implementation of that part with templates, as we do with the rest.

PR #150, opened 2019-07-16
[9.0] [FIX] Product Variant Configurator: product pricelist

ThomasBinsfeld:


pedrobaeza:
Can you explain changes? Is this applicable to upper versions?

ThomasBinsfeld:
@pedrobaeza I'm not an expert of this module, but i seems it rewrites a lot of Odoo code. I just fixed the code by comparing it to [Odoo code](https://github.com/odoo/odoo/blob/9.0/addons/product/pricelist.py#L235).

Also, the unit tests were wrong, IMO. They were using _formula_ price lists with `percent_price`, which is impossible from the interface. A _formula_ price list on which you want to define a percentage of discount uses the _price_discount_ field. That's why the unit tests were unable to detect the bug.

pedrobaeza:
Well, the module was mostly rewritten by your colleague @zakiuu, so if he is available, he can give an opinion.

pedrobaeza:
Sorry, wrong button

ThomasBinsfeld:
> Is this applicable to upper versions?

@pedrobaeza It looks OK in the v12 migration PR.

pedrobaeza (approved):
OK, no blocking by my part. I don't have now any customer in 9.0

chrisdec (approved):
Functional tests

PR #158, opened 2019-10-25
add missing configuration doc

zeroheure:
I've left fields and config values typography as I don't yet understand OCA typo codes.


pedrobaeza (approved):

zeroheure:
@pedrobaeza I'm rewriting the doc, which is very confusing (looks like everyone added some paragraphs when adding functions :-). Not yet finished.

zeroheure:
Finished now. The last commit explains conventions I've used.

PR #162, opened 2020-01-08
[ADD] product_variant_sale_order_route

p-tombez:
Add a module to specify a specific route on product variants to be used on SO lines


product_variant_sale_order_route/models/product.py
```diff
@@ -0,0 +1,15 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+
+from odoo import models, fields
+
+
+class Product(models.Model):
+    _inherit = 'product.product'
+
+    sale_order_line_route_id = fields.Many2one(
```
grindtildeath, line 11: This must be a M2m to allow different routes for the same variant.
Moreover, I'm wondering if it shouldn't be named `route_ids` in order to obfuscate `product.template`'s `route_ids`.

product_variant_sale_order_route/models/sale_order_line.py
```diff
@@ -0,0 +1,17 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+
+from odoo import models
+
+
+class SaleOrderLine(models.Model):
+    _inherit = 'sale.order.line'
+
+    def product_id_change(self):
```
grindtildeath, line 11: This doesn't look right. IMO we need to redefine the domain applied to `route_id` dynamically according to the product.

grindtildeath (changes requested):
We'll also need a test here..

product_variant_sale_order_route/models/product.py
grindtildeath: After checking the customer requirements again, as we only need ONE route per variant, let's keep it simple as it is, instead of complicating things :wink:

product_variant_sale_order_route/models/sale_order_line.py
grindtildeath: As we want a M2o on product.product, the onchange is right :wink:

product_variant_sale_order_route/views/product_view.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" ?>
+<odoo>
+
+    <record model="ir.ui.view" id="product_normal_form_view">
+        <field name="name">product_normal_form_view</field>
+        <field name="model">product.product</field>
+        <field name="inherit_id" ref="product.product_normal_form_view" />
+        <field name="arch" type="xml">
+            <field name="route_ids" position="after">
```
grindtildeath, line 9: IMO just set route_ids as invisible on the product.product form (so that it won't update the product.template) and we should be fine

Tonow-c2c (approved):
Just code review
for functional things @grindtildeath have truly better advice

gurneyalex (approved):
Travis red build is not related to this PR.

PR #164, opened 2020-01-20
[FIX] compute_list_price

oscars8a:
Hi, I'm Oscar.

I have found this problem with the unit  of measure:
In a new DB, I have only installed sale_management and product_variant_sale_price modules, I have created a unit of measure "Three months" with factor_inv = 3.0.
The problem is that when I add a product in a sale.order.line, the calculation of price_unit is wrong.
The calculation is price = list_price * factor_inv, but if I install product_variant_sale_price this calculation is executed twice: price = (list_price * factor_inv)² .

This PR only remove the suspects lines of code.

Thanks a lot.

The following gif shows the problem.

error_in_product_variant_oca

### DIFF
--- a/product_variant_sale_price/models/product_product.py
+++ b/product_variant_sale_price/models/product_product.py
@@ -44,12 +44,8 @@ class ProductProduct(models.Model):
 
     @api.multi
     def _compute_list_price(self):
-        uom_model = self.env['uom.uom']
         for product in self:
             price = product.fix_price or product.product_tmpl_id.list_price
-            if 'uom' in self.env.context:
-                price = product.uom_id._compute_price(
-                    price, uom_model.browse(self.env.context['uom']))
             product.list_price = price
 
     @api.multi


JorgeQuinteros (approved):

spoerp:
This indeed is a major issue, and upon further testing, I found that this PR does indeed fixes the incorrect pricing with different UOMs

oscars8a:
> This indeed is a major issue, and upon further testing, I found that this PR does indeed fixes the incorrect pricing with different UOMs

Thanks @spoerp maybe you can do a review for this PR

@pedrobaeza What is the way we have to follow?

Thanks a lot

pedrobaeza:
Get reviewers (one should be from the repo PSC), and then it will be merged.

oscars8a:
> Get reviewers (one should be from the repo PSC), and then it will be merged.

Excuse me, @pedrobaeza , Who is the PSC from this Repo, How can I know it?

Thanks a lot.

pedrobaeza:
https://odoo-community.org/psc-teams/product-27

oscars8a:
Hi @rousseldenis, Can you review this PR as PSC?

Thanks a lot.

product_variant_sale_price/models/product_product.py
```diff
@@ -44,12 +44,8 @@ def _compute_lst_price(self):
 
     @api.multi
     def _compute_list_price(self):
-        uom_model = self.env['uom.uom']
         for product in self:
             price = product.fix_price or product.product_tmpl_id.list_price
```
rousseldenis, line 48: Maybe the good fix should do an if/else with that line?

rousseldenis: Furthermore, I would load once uom if it is in context before the loop

spoerp:
@oscars8a @pedrobaeza  I am going to create a PR for Migration of this module to v13. It will also resolve the issue with lst_price that loses the price coming from attributes' price_extra after installing the module.

PR #174, opened 2020-06-04
[MIG] product_variant_specific_tax Migration to 13.0

GiacoJona:


pedrobaeza:
I'm afraid this still doesn't contain the commit history. If the target module is being migrated from other branch that is not 12.0, you need to change the command in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-13.0 to match the source version.

GiacoJona:
It's from branch 11.0, i change the source version in the command but i guess i did something wrong but i don't know what, yet.

PR #195, opened 2020-11-29
[12.0][FIX] if a product already have a code do not change it

sebastienbeau:
When modify a attribute value (fixing a typo for example).
The onchange will change the code of the attribute.
This may generate a lot of mistake as user can change the code of an attribute value by mistake and set a wrong value

### DIFF
--- a/product_variant_default_code/models/product.py
+++ b/product_variant_default_code/models/product.py
@@ -203,7 +203,7 @@ class ProductAttributeValue(models.Model):
 
     @api.onchange('name')
     def onchange_name(self):
-        if self.name:
+        if self.name and not self.code:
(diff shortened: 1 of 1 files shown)


sebastienbeau:
@pedrobaeza @bealdav @alfredoavanzosc

bealdav (approved):

pedrobaeza (changes requested):
I think this is a change of behavior, as I find legit cases both on previous code and new one, depending on the users. If you add this, you should do it configurable (via a system parameter for example).

bealdav:
Please @sebastienbeau.

Could you close or any option please ?

PR #204, opened 2021-01-27
[MIG] sale_variant_configurator: Migration to 13.0

chienandalu:
cc @Tecnativa TT27133


chienandalu:
@pedrobaeza @Yajo Please, review

sale_variant_configurator/models/sale_order.py
```diff
@@ -73,7 +89,7 @@ def _onchange_product_tmpl_id_configurator(self):
         self.tax_id = fpos.map_tax(taxes) if fpos else taxes
         product_tmpl = self.product_tmpl_id.with_context(
             lang=self.order_id.partner_id.lang,
-            partner=self.order_id.partner_id.id,
+            partner=self.order_id.partner_id,
```
yajo: It is very uncommon to put a recordset in the context, because usually contexts can be passed in by RPC calls, which only can use basic data types. Are you sure this is correct/needed? 🤔

sale_variant_configurator/models/sale_order.py
```diff
@@ -113,7 +129,7 @@ def _update_price_configurator(self):
             return
         product_tmpl = self.product_tmpl_id.with_context(
             lang=self.order_id.partner_id.lang,
-            partner=self.order_id.partner_id.id,
+            partner=self.order_id.partner_id,
```
yajo: Same concern as above.

yajo (approved):
I see one oddity, but it's probably OK, so 👍🏼 

Check bots please

sale_variant_configurator/models/sale_order.py
chienandalu: The thing is I was having a  `can't mix apples and oranges` warning

chienandalu: A core example on it: https://github.com/odoo/odoo/blob/13.0/addons/sale_product_configurator/controllers/main.py#L22

pedrobaeza: This context is server-side, so no risk.

yajo: OK, I assumed it'd be like that, but it was weird anyway so I asked. Thanks for the pointers 😊

pedrobaeza (approved):

PR #205, opened 2021-01-28
[13.0][ADD] product_variant_attribute_tax: New module

ernestotejeda:
Cc @Tecnativa TT27125

This module allows to set taxes on the product attribute values.


product_variant_attribute_tax/models/product_attribute_value.py
```diff
@@ -0,0 +1,10 @@
+# Copyright 2016-2017 Tecnativa - Pedro M. Baeza
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+
+from odoo import fields, models
+
+
+class ProductAttributeValue(models.Model):
+    _inherit = "product.attribute.value"
+
+    tax_ids = fields.Many2many(comodel_name="account.tax", string="Taxes")
```
rousseldenis, line 10: Shouldn't code be aligned on Odoo for products and add a 'supplier_tax_ids' ?

pedrobaeza: The idea here is to have everything together for simpler UI, and the tax type (sale/purchase), defines to which apply on subsequent modules (for now, only `sale_product_attribute_tax`).

product_variant_attribute_tax/__manifest__.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2016-2017 Tecnativa - Pedro M. Baeza
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Taxes on product attribute values",
+    "summary": "Set taxes on the product attribute values",
+    "author": "Tecnativa, Odoo Community Association (OCA)",
+    "website": "http://www.tecnativa.com",
```
pedrobaeza: Change this to GitHub repo URL

pedrobaeza (approved):

product_variant_attribute_tax/__manifest__.py
ernestotejeda: Done

pedrobaeza:
@rousseldenis is it OK for you?

product_variant_attribute_tax/models/product_attribute_value.py
rousseldenis: What I wanted to say, as both are defined at account module level(and so dependency is the same level), why not doing the same here.     https://github.com/odoo/odoo/blob/fb01006342c6d6fd902e13a5f4c2642c87fe452d/addons/account/models/product.py#L27

And this module is not talking about sale or purchase but just product

pedrobaeza: Yeah, I got you, but from UX perspective, I still prefer to have only one field to fill the involved taxes (being for purchase or for sale), and then filter them on operation. Two reasons for this:

- As it's an editable table, having one column instead of 2 is better for not having high widths on both columns, as some attributes can be unbalanced (only sales taxes or purchase taxes).
- On selection, don't have to look for which taxes are of one type or another. All are selected at the same time.

Do you get me?

joao-p-marques (approved):
Code review :+1: LGTM 🙂

HaraldPanten (approved):

PR #209, opened 2021-02-07
[13.0][IMP] product_variant_configurator: Use Many2oneReference + index

pedrobaeza:
Use the new type of field `Many2oneReference` for favoring further features in it (make use of caching system) and add index on this field.

@Tecnativa TT27113


chienandalu (approved):

ThomasBinsfeld (approved):
:ok_hand:

PR #211, opened 2021-02-18
[FIX] product_variant_default_code: fix onchange code value

chienandalu:
When an attribute code is already set, we don't wan't to change it by
chance if we set a new name in the value. The onchange should be just a
helper to fill the initial value but it should be manually changed if
needed later on.

cc @Tecnativa TT28258


pedrobaeza:
Travis is red

chienandalu:
Ok, tests fixed

pedrobaeza (approved):

victoralmau (approved):

PR #212, opened 2021-02-26
[14.0][FIX] product_variant_default_code: Fix tests

joao-p-marques:
Tests can conflict with constriants added by other modules for the `default_code`
We should create different product templates only if we can be sure they will compute different `default_codes`, otherwise they might trigger the `unique(default_code)` constraint, if introduced by another module.

@Tecnativa
TT28423

ping @pedrobaeza @victoralmau


pedrobaeza (approved):

pedrobaeza:
Fast-tracking these test fix:

/ocabot merge patch

PR #213, opened 2021-03-04
[13.0] [FIX] product_variant_configurator: attribute readonly

ThomasBinsfeld:
Attribute must be readonly. Probably a bug introduced during the migration.


pedrobaeza:
It was explicitly changed in the 13.0 migration by @HviorForgeFlow. Héctor, do you remember why?

pedrobaeza (approved):
This must be indeed True

ThomasBinsfeld:
Thanks!

PR #215, opened 2021-05-03
[13.0][ADD] product_variant_change_attribute_value

TDu:
With standard Odoo there is no way to change the attribute values assigned
to a product variant.

This module adds a wizard that can be access through the product variant
tree view context action menu.

The wizard allows to change or remove attribute values on all product
variant selected.

When an attribute value is removed from all variant of a template,
it will also be removed from the configuration of the related product template.


sebalix:
@TDu I open https://github.com/OCA/product-variant/pull/216 instead

PR #217, opened 2021-05-24
[14.0][MIG] sale_order_line_variant_description

SimoneVagile:

PR #218, opened 2021-06-04
[14.0] move to GitHub actions

sbidoul:
As there are travis issues in https://github.com/OCA/product-variant/pull/217, I propose to use this repo as the next test for GitHub actions CI.

mis-builder (14.0) and pim (all branches) are already configured that way, as well as all shoinvader repos.

Things to note:
- compared to travis/mqt, these it should run the exact same tests, so no surprise expected, but
- pay particular attention to false positives (i.e. read the build logs to detect anomalies)
- there are additional checks for license and development status compatibility


ivantodorovich (approved):
This is great :100:

rousseldenis (approved):

sbidoul:
Hm, I almost forgot. If we remove .travis.yml, we lose runbot.

ivantodorovich:
Is it possible to run `manifestoo` on pre-commit instead? I'm thinking it could speed up the process if the developer sees the warning before even creating the PR.

Ideally it'd be great to do the same with `oca-addon-gen-readme`, to reduce the number of administrative commits

sbidoul:
> Is it possible to run `manifestoo` on pre-commit instead? I'm thinking it could speed up the process if the developer sees the warning before even creating the PR.

The difficulty is that it requires network access to download the direct dependencies and inspect their manifest. The ci know how to download and install dependencies. Bringing that knowledge to pre-commit is more difficult and generally pre-commit is used for local, static checks. 

> Ideally it'd be great to do the same with oca-addon-gen-readme, to reduce the number of administrative commits

Maybe. I think the idea came up before but I don't remember exactly why it was not implemented. Maybe due to the additional risk of merge conflicts ?

PR #220, opened 2021-07-13
[13.0][ADD] product_variant_list_pricelist_price

JordiMForgeFlow:
This module allows displaying the sales price of a product variant for each active pricelist in the system as optional columns in the product variant list view.

The public price of the variant is also hidden from the view.


JordiMForgeFlow:
The functionality is being extended to product template level, so the MR has been readapted here: https://github.com/OCA/product-attribute/pull/883

PR #216, opened 2021-05-03
[13.0][ADD] product_variant_change_attribute_value

sebalix:
Same PR than https://github.com/OCA/product-variant/pull/215 but based on an older commit (ping @TDu).

With standard Odoo there is no way to change the attribute values assigned
to a product variant.

This module adds a wizard that can be access through the product variant
tree view context action menu.

The wizard allows to change or remove attribute values on all product
variant selected.

When an attribute value is removed from all variant of a template,
it will also be removed from the configuration of the related product template.

Ref. 2090

PR #222, opened 2021-08-09
[13.0][ADD] product_variant_change_attribute_value

simahawk:
Replaces https://github.com/OCA/product-variant/pull/216

I'll squash some commits after approval.

### DIFF
--- a/product_variant_change_attribute_value/tests/test_product_variant_change_attribute_value.py
+++ b/product_variant_change_attribute_value/tests/test_product_variant_change_attribute_value.py
@@ -79,7 +79,7 @@ class TestProductVariantChangeAttributeValue(common.SavepointCase):
 
         wiz = self.wizard.with_context(default_res_ids=self.variants).create({})
         self._change_action(wiz, self.steel, "delete")
-        wiz.action_change_attributes()
+        wiz.action_apply()
 
         self.assertFalse(self.is_value_on_variant(self.variant_1, self.steel))
         self.assertFalse(
@@ -92,7 +92,7 @@ class TestProductVariantChangeAttributeValue(common.SavepointCase):
 
         wiz = self.wizard.with_context(default_res_ids=self.variants).create({})
         self._change_action(wiz, self.white, "replace", self.pink)
-        wiz.action_change_attributes()
+        wiz.action_apply()
 
         self.assertFalse(self.is_value_on_variant(self.variant_1, self.white))
         self.assertTrue(self.is_value_on_variant(self.variant_1, self.pink))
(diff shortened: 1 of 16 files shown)


product_variant_change_attribute_value/wizards/product_variant_attribute_value_action.py
```diff
@@ -0,0 +1,66 @@
...
+            WHERE
+                attribute_id IN %(ids)s
+            GROUP BY
+                attribute_id
+        """
+        ids = tuple(self.mapped("attribute_id").ids)
+        if not ids:
+            self.update({"selectable_attribute_value_ids": False})
+            return
+        self.env.cr.execute(query, dict(ids=ids))
```
grindtildeath, line 62:
```suggestion
        self.env["product.attribute.value"].flush(["attribute_id"])
        self.env.cr.execute(query, dict(ids=ids))
```

product_variant_change_attribute_value/wizards/product_variant_attribute_value_wizard.py
```diff
@@ -0,0 +1,209 @@
...
+
+import psycopg2
+
+from odoo import _, api, fields, models
+from odoo.exceptions import UserError
+
+
+class VariantAttributeValueWizard(models.TransientModel):
+    _name = "variant.attribute.value.wizard"
+    _description = "Wizard to change attriubtes on product variants"
```
grindtildeath:
```suggestion
    _description = "Wizard to change attributes on product variants"
```

grindtildeath (commented):
I couldn't test on runbot yet but LG

grindtildeath (commented):
After testing on runbot, it looks like it's quite easy to mess things up by deleting attribute values, and we're not able to recreate these on an existing product.
I'd suggest to state limitations in the readme, and restrict access to the wizard to a dedicated user group.

TDu (approved):

mmequignon (changes requested):
https://user-images.githubusercontent.com/13200247/143858116-c6e984e1-c35a-4fc5-a1cf-2bdf116fcf59.mp4

There's a nasty bug here.

product_variant_change_attribute_value/tests/test_product_variant_change_attribute_value.py
```diff
@@ -88,9 +89,10 @@ def test_fields(self):
     def test_actions_field_filter(self):
         wiz = self._get_wiz()
         self.assertEqual(len(wiz.attributes_action_ids), len(wiz.attribute_value_ids))
-        wiz.filter_attribute_id = self.legs
+        with Form(wiz) as res:
+            res.filter_attribute_id = self.legs
         self.assertEqual(
-            len(wiz.attributes_action_ids),
+            len(res.attributes_action_ids),
```
TDu, line 95: Should the assert not be in the context manager ?

mmequignon: Maybe. Probably. I have to switch on something else. I'll come back to this later :smile:

simahawk:
@mmequignon do you think we can finalize this (maybe w/ improvements from 14)?

PR #223, opened 2021-09-13
[14.0][IMP] product_variant_default_code: prefix_code as default_code by default

Kev-Roche:
By default, when a product have variants, its default_code is False. No default_code can be disturbing in some products' views.
Here, we change it by putting the prefix_code as default_code by default. 

The change result in view :
Capture d’écran de 2021-09-13 19-53-34
Capture d’écran de 2021-09-13 19-52-50

### DIFF
--- a/product_variant_default_code/models/config_settings.py
+++ b/product_variant_default_code/models/config_settings.py
@@ -17,3 +17,9 @@ class BaseConfiguration(models.TransientModel):
         implied_group="product_variant_default_code"
         ".group_product_default_code_manual_mask",
     )
+
+    prefix_as_default_code = fields.Boolean(
+        string="Reference Prefix as default Reference",
+        default=False,
+        config_parameter="product_variant_default_code.prefix_as_default_code",
+    )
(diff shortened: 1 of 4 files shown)


pedrobaeza:
I don't get why we would want this behavior by default...

sebastienbeau:
Hi Pedro, this change is based on customer feedback regarding the ergonomic of this module
- first it's really confusing for them to not see the code of the product template when it have variant. Indeed for them the prefix code is the code of the template
- also if you do a search with the prefix_code you will not find the product

For sure we can add an extra column and inherit search behaviour (everywhere) but just inheriting the default_code of the template solve both issue and provide the best UI.

I will be interested by other feedback

pedrobaeza:
I see this more as a workaround than a perfect UI solution, and depending on the prefix_code, it can be even more confusing IMO.

sebastienbeau:
If you prefer we can move this in a small module or adding an option in the current module

pedrobaeza:
Yeah, an opt-in option disabled by default is OK for me.

Kev-Roche:
@pedrobaeza, @sebastienbeau , I added the option disabled by default

pedrobaeza (approved):
Thanks for doing it optional

legalsylvain (approved):
thanks. LGTM.

PR #227, opened 2021-11-17
[14.0][FIX] product_variant_default_code : reference_mask and default_code not updated

Kev-Roche:
Some problems occur when using the variant default code. In the following cases (with corresponding tests), reference_mask (product.template) and default_code (variants/product.product) are not updated :
1) If the code_prefix changes : test_12_prefix_change
2) If we add a new variant attribute : test_13_new_attribute
3) If we rename a variant attribute : test_14_rename_attribute
4) If we change the variant attributes’ sequence : test_15_sequence_change

### DIFF
--- a/product_variant_default_code/models/product.py
+++ b/product_variant_default_code/models/product.py
@@ -119,8 +119,10 @@ class ProductTemplate(models.Model):
     def _compute_reference_mask(self):
         automask = self.is_automask()
         for rec in self:
-            if automask or rec.reference_mask == "":
+            if automask or not rec.reference_mask:
                 rec.reference_mask = rec._get_default_mask()
+            elif not automask and rec.code_prefix:
+                rec.reference_mask = rec.code_prefix + rec.reference_mask
 
     def _inverse_reference_mask(self):
         self._compute_reference_mask()
--- a/product_variant_default_code/models/product.py
+++ b/product_variant_default_code/models/product.py
@@ -4,6 +4,7 @@
 # Copyright 2017 Tecnativa - David Vidal
 # Copyright 2018 Avanzosc S.L. - Daniel Campos
 # Copyright 2020 Tecnativa - João Marques
+# Copyright 2021 Akretion - Kévin Roche
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 import re
@@ -40,34 +41,6 @@ def sanitize_reference_mask(product, mask):
         )
 
 
-def get_rendered_default_code(product, mask):
-    product_attrs = defaultdict(str)
-    reference_mask = ReferenceMask(mask)
-    main_lang = product.product_tmpl_id._guess_main_lang()
-    for attr in product.product_template_attribute_value_ids:
-        value = attr.product_attribute_value_id
-        attr_name = value.attribute_id.with_context(lang=main_lang).name
-        if value.attribute_id.code:
-            product_attrs[attr_name] += value.attribute_id.code
-        if value.code:
-            product_attrs[attr_name] += value.code
-    all_attrs = extract_token(mask)
-    missing_attrs = all_attrs - set(product_attrs.keys())
-    missing = dict.fromkeys(
-        missing_attrs,
-        product.env["ir.config_parameter"]
-        .sudo()
-        .get_param("default_reference_missing_placeholder"),
-    )
-    product_attrs.update(missing)
-    default_code = reference_mask.safe_substitute(product_attrs)
-    return default_code
-
-
-def render_default_code(product, mask):
-    product.default_code = get_rendered_default_code(product, mask)
-
-
 class ProductTemplate(models.Model):
     _inherit = "product.template"
 
@@ -78,6 +51,9 @@ class ProductTemplate(models.Model):
     reference_mask = fields.Char(
         string="Variant reference mask",
         copy=False,
+        store=True,
+        compute="_compute_reference_mask",
+        inverse="_inverse_reference_mask",
         help="Reference mask for building internal references of a "
         "variant generated from this template.\n"
         "Example:\n"
@@ -100,12 +76,62 @@ class ProductTemplate(models.Model):
         "attribute name",
     )
 
+    variant_default_code_error = fields.Text(
+        compute="_compute_variant_default_code_error"
+    )
+
+    def is_automask(self):
+        return bool(
+            not self.user_has_groups(
+                "product_variant_default_code.group_product_default_code_manual_mask"
+            )
+        )
+
+    @api.depends(
+        "code_prefix",
+        "attribute_line_ids.value_ids",
+        "attribute_line_ids.value_ids.code",
+        "attribute_line_ids.value_ids.name",
+    )
+    def _compute_variant_default_code_error(self):
+        automask = self.is_automask()
+        for rec in self:
+            error_txt = ""
+            if not rec.code_prefix and automask:
+                error_txt += "Reference Prefix is missing.\n"
+            invalid_values = self.attribute_line_ids.value_ids.filtered(
+                lambda s: not s.code
+            )
+            if invalid_values:
+                error_txt += (
+                    "Following attribute value have an empty code :\n- "
+                    + "\n- ".join(invalid_values.mapped("name"))
+                )
+            if error_txt:
+                error_txt = "Default Code can not be computed.\n" + error_txt
+            rec.variant_default_code_error = error_txt
+
+    @api.depends(
+        "code_prefix",
+        "attribute_line_ids",
+        "attribute_line_ids.attribute_id.name",
+    )
+    def _compute_reference_mask(self):
+        automask = self.is_automask()
+        for rec in self:
+            if automask or rec.reference_mask == "":
+                rec.reference_mask = rec._get_default_mask()
+
+    def _inverse_reference_mask(self):
+        self._compute_reference_mask()
+
     def _get_default_mask(self):
         attribute_names = []
         default_reference_separator = (
             self.env["ir.config_parameter"]
             .sudo()
             .get_param("default_reference_separator")
+            or ""
         )
         # Get the attribute name in the main lang format, otherwise we could not
         # match mask with the proper values
@@ -133,30 +159,6 @@ class ProductTemplate(models.Model):
         elif vals.get("reference_mask"):
             sanitize_reference_mask(product, vals["reference_mask"])
         return super(ProductTemplate, self).create(vals)
-
-    def write(self, vals):
-        with_variants = self.env["product.template"]
-        product_obj = self.env["product.product"]
-        if (
-            "reference_mask" in vals
-            and not vals["reference_mask"]
-            or not self.user_has_groups(
-                "product_variant_default_code.group_product_default_code_manual_mask"
-            )
-        ):
-            with_variants = self.filtered("attribute_line_ids")
-            for template in with_variants:
-                new_dict = dict(vals)
-                new_dict["reference_mask"] = template._get_default_mask()
-                super(ProductTemplate, template).write(new_dict)
-        super(ProductTemplate, self - with_variants).write(vals)
-        if vals.get("reference_mask"):
-            cond = [("product_tmpl_id", "=", self.id), ("manual_code", "=", False)]
-            products = product_obj.search(cond)
-            for product in products:
-                if product.reference_mask:
-                    render_default_code(product, product.reference_mask)
-        return True
 
     @api.model
     def _guess_main_lang(self):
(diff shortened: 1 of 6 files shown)


product_variant_default_code/models/product.py
```diff
@@ -211,43 +213,21 @@ class ProductAttribute(models.Model):
...
 
 
 class ProductAttributeValue(models.Model):
     _inherit = "product.attribute.value"
 
-    @api.onchange("name")
-    def onchange_name(self):
-        """Set a default code for the value"""
+    @api.onchange("code", "name")
+    def onchange_default_code_values(self):
```
sebastienbeau, line 228: Why did you rename this onchange (normally we name in onchange_name because the field that have been changed is "name")

CasVissers-360ERP (approved):
Functional test, this fixes our issue with the default_code not updating when the code_prefix was changed. Other improvements seem good to me as well.

CasVissers-360ERP:
@Kev-Roche 
The change can give errors when using in combination with OCA product reference unique:
psycopg2.errors.UniqueViolation: duplicate key value violates unique constraint "product_product_default_code_uniq"
DETAIL:  Key (default_code)=(False False) already exists. 

Not sure if this is new to this update.

product_variant_default_code/models/product.py
```diff
@@ -264,4 +244,13 @@ def write(self, vals):
...
+
+    code = fields.Char(
+        string="Attribute Value Code",
+    )
+
+    @api.model
+    def create(self, vals):
+        if "code" not in vals:
+            vals["code"] = vals.get("name", "")[0:2]
+        return super(ProductAttributeValue, self).create(vals)
```
sebastienbeau:
```suggestion
        return super().create(vals)
```

product_variant_default_code/tests/test_variant_default_code.py
```diff
@@ -208,12 +221,93 @@ def test_09_code_change_propagation(self):
...
+                + product.product_template_attribute_value_ids.filtered(
+                    lambda x: x.product_attribute_value_id.attribute_id == self.attr3
+                ).name[0:2]
+            )
+            self.assertEqual(product.default_code, expected_code)
+
+    def test_14_rename_attribute(self):
+        self.assertEqual(self.template1.reference_mask, "[TSize]-[TColor]")
+        self.attr1.name = "TNewSize"
+        self.attr1.onchange_default_code_attribute()
```
sebastienbeau: in your test you should no call the onchange, writing on name should be enough

product_variant_default_code/models/product.py
```diff
@@ -100,12 +104,34 @@ class ProductTemplate(models.Model):
...
+    def _compute_reference_mask(self):
+        automask = bool(
+            not self.user_has_groups(
+                "product_variant_default_code.group_product_default_code_manual_mask"
+            )
+        )
+        for rec in self:
+            if automask or rec.reference_mask == "":
+                rec.reference_mask = rec._get_default_mask()
+            rec._apply_reference_mask()
```
sebastienbeau: instead of calling manually the _apply_reference_mask.
It will be better to add a computed field on the default_code of the product_product.

somethink like
```
@api.depends(
    "product_tmpl_id.reference_mask",
    "product_template_attribute_value_ids.product_attribute_value_id.code",
     "product_template_attribute_value_ids.attribute_id.code",
)
def _compute_default_code(self):
   ...

```

sebastienbeau (changes requested):

product_variant_default_code/tests/test_variant_default_code.py
```diff
@@ -208,12 +221,93 @@ def test_09_code_change_propagation(self):
...
+                self.template1.code_prefix
+                + product.product_template_attribute_value_ids.filtered(
+                    lambda x: x.product_attribute_value_id.attribute_id == self.attr1
+                ).name[0:2]
+                + "-"
+                + product.product_template_attribute_value_ids.filtered(
+                    lambda x: x.product_attribute_value_id.attribute_id == self.attr2
+                ).name[0:2]
+            )
+            product.flush()
```
sebastienbeau: remove flush

product_variant_default_code/tests/test_variant_default_code.py
```diff
@@ -208,12 +221,93 @@ def test_09_code_change_propagation(self):
...
+        self.attr1.onchange_default_code_attribute()
+        self.assertEqual(self.template1.reference_mask, "[TNewSize]-[TColor]")
+
+    def test_15_sequence_change(self):
+        self.assertEqual(self.template1.reference_mask, "[TSize]-[TColor]")
+        self.attr1.sequence = 1
+        self.attr2.sequence = 0
+        self.template1.reference_mask = "[TColor]-[TSize]"
+        self.template1.write({"name": "New"})
+        self.assertEqual(self.template1.reference_mask, "[TColor]-[TSize]")
```
sebastienbeau: TODO
- add a test where you change a code of a product.attribute
- add a test where you change a code of a product.attribute.value

sebastienbeau (changes requested):

product_variant_default_code/models/product.py
Kev-Roche: @sebastienbeau, thanks for your suggestions, I apply them.

product_variant_default_code/tests/test_variant_default_code.py
Kev-Roche: Tests 9 and 10 already test them, I just add the code change of an active product.attribute.value in test 9

Kev-Roche:
@CasVissers-360ERP @sebastienbeau thank you for your relevant reviews, I improve the PR accordingly.
I m not sure about the field "manual_code", which is now activate or not by the user and false by default. I think it is a good option since this module is mostly usefull in automatic / not manual mode. what do you think about it?

product_variant_default_code/models/product.py
```diff
@@ -100,12 +76,32 @@ class ProductTemplate(models.Model):
         "attribute name",
     )
 
+    @api.depends(
+        "code_prefix",
+        "attribute_line_ids",
+        "attribute_line_ids.attribute_id.name",
+        "attribute_line_ids.attribute_id.code",
```
sebastienbeau: no sure that we need to depend on  "attribute_line_ids.attribute_id.code" here

product_variant_default_code/models/product.py
```diff
@@ -186,18 +158,43 @@ def _compute_default_code(self):
...
+                all_attrs = extract_token(rec.product_tmpl_id.reference_mask)
+                missing_attrs = all_attrs - set(product_attrs.keys())
+                missing = dict.fromkeys(
+                    missing_attrs,
+                    rec.env["ir.config_parameter"]
+                    .sudo()
+                    .get_param("default_reference_missing_placeholder"),
+                )
+                product_attrs.update(missing)
+                default_code = reference_mask.safe_substitute(product_attrs)
```
sebastienbeau: can you move this code in a method _build_default_code or _generate_default_code.
But this method is normal odoo method of the class product.product (not like before with the method get_rendered_default_code)

product_variant_default_code/tests/test_variant_default_code.py
sebastienbeau: ok

product_variant_default_code/models/product.py
```diff
@@ -186,18 +158,43 @@ def _compute_default_code(self):
...
-
-    @api.model
-    def create(self, vals):
-        product = super(ProductProduct, self).create(vals)
-        if product.reference_mask:
-            render_default_code(product, product.reference_mask)
-        return product
+    manual_code = fields.Boolean(string="Manual Reference", default=False)
+    default_code = fields.Char(
+        compute="_compute_default_code", readonly=False, store=True
```
sebastienbeau: maybe you should add an inverse method so if you fill manually the default_code it tick the manual_code to True (like the onchange_default_code

product_variant_default_code/models/product.py
```diff
@@ -186,18 +158,43 @@ def _compute_default_code(self):
...
+    )
 
-    @api.onchange("default_code")
-    def onchange_default_code(self):
-        self.manual_code = bool(self.default_code)
+    @api.depends(
+        "product_tmpl_id.reference_mask",
+        "product_template_attribute_value_ids.attribute_id.code",
```
sebastienbeau, line 203: maybe you do not need to depend on 
product_template_attribute_value_ids.attribute_id.sequence
product_template_attribute_value_ids.attribute_id.name
product_template_attribute_value_ids.product_attribute_value_id.name
as this dependency will change the default mask

as this

product_variant_default_code/models/product.py
Kev-Roche: Indeed, not necessary

Kev-Roche:
cc @sebastienbeau, @CasVissers-360ERP, @joao-p-marques

sebastienbeau:
@CasVissers-360ERP  I successfully reproduce the bug with the default code uniq. I am going to fix it.
The issue is that the default_code is set to False (or "") instead of None

sebastienbeau:
After looking twice, we have an conceptual issue of compatibility between product_variant_default_code and product_code_unique.

Indeed if some information is missing (like the code_prefix or an attribute value) the module product_variant_default_code will generate "useless" and "wrong" default code with a lot of chance to be identical (if you have 2 attribute value without code or 2 product without prefix => you have duplicated default_code)

As generating incomplete default_code is for me useless (and I prefer to have nothing then something wrong).
Maybe it's better to not generate the default_code if something is missing ?
@pedrobaeza @CasVissers-360ERP @bealdav

CasVissers-360ERP:
@sebastienbeau 
>> Maybe it's better to not generate the default_code if something is missing ?

For me that would work. Maybe add a logger warning or something so it's easier to debug while the SKU generation is failing?

sebastienbeau:
Thanks for your feedback, what we can do is showing a warning in the tab "variants" of the product template 

Something like 


image


And we can add a filter in the search view to get all template with missing information

CasVissers-360ERP:
That's very clean, the filter would be a bit overkill for me. Administrators can use the advanced filter to filter them.

pedrobaeza:
I agree to not generate the code if something is missing.

sebastienbeau:
@Kev-Roche can you do it

sebastienbeau:
@pedrobaeza @CasVissers-360ERP
After thinking twice what do you think about making the field code (of product.attribute.value) mandatory in this module so we do not have any error message to add ?

Then we can have extra module if we want to have the prefix_code (of product.template) and or the code (of product.attribute) required.

pedrobaeza:
I don't think so, as there can be a lot of ways to create the product, and make them to fail due to that hard constraint is too much IMO.

CasVissers-360ERP:
> I don't think so, as there can be a lot of ways to create the product, and make them to fail due to that hard constraint is too much IMO.

I agree with that.

Kev-Roche:
To resume which changes are required :
- No compute of the default_code and warning message in case of 
1) missing prefix_code 
2) missing product.attribute.value code.
3) if the product.attribute.value code is not unique for each attribute (duplicate). 
- In manual mode : only attribute_value code is needed (not the prefix_code). 

Are you agreed with that?

pedrobaeza:
I think the point 3 is out of scope of this module.

Kev-Roche:
cc @sebastienbeau, @pedrobaeza, @CasVissers-360ERP

product_variant_default_code/models/product.py
```diff
@@ -76,19 +76,51 @@ class ProductTemplate(models.Model):
...
+        "attribute_line_ids.value_ids.code",
+        "attribute_line_ids.value_ids.name",
+    )
+    def _compute_variant_default_code_error(self):
+        if self.is_automask():
+            for rec in self:
+                error_txt = ""
+                if not rec.code_prefix and self.is_automask():
+                    error_txt += "Reference Prefix is missing.\n"
+                values = self.attribute_line_ids.value_ids
```
sebastienbeau: Maybe we can simplify the code by doing something like
```
invalid_values = self.attribute_line_ids.value_ids.filtered(lambda s:  not s.code)
if invalid_values:
    error_txt += "Following attribute value have an empty code : \n-" + "\n-".join(invalid_values.mapped("name"))
```

product_variant_default_code/models/product.py
```diff
@@ -76,19 +76,51 @@ class ProductTemplate(models.Model):
...
+                if not all(values.mapped("code")):
+                    error_txt += "Some attribute codes are missing:\n"
+                    for value in values:
+                        if not value.code:
+                            error_txt += "- The code of the attribute value"
+                            error_txt += " '{}' is missing\n".format(value.name)
+                if error_txt:
+                    error_txt = "".join(
```
sebastienbeau: error_txt = "Default Code can not be computed.\n" + error_txt

product_variant_default_code/models/product.py
```diff
@@ -76,19 +76,51 @@ class ProductTemplate(models.Model):
...
-        "code_prefix",
-        "attribute_line_ids",
-        "attribute_line_ids.attribute_id.name",
+    variant_default_code_error = fields.Text(
+        compute="_compute_variant_default_code_error"
     )
-    def _compute_reference_mask(self):
+
+    def is_automask(self):
         automask = bool(
```
sebastienbeau: the variable automask is not necessary you can directly return bool(...)

product_variant_default_code/models/product.py
```diff
@@ -76,19 +76,51 @@ class ProductTemplate(models.Model):
...
+
+    @api.depends(
+        "code_prefix",
+        "attribute_line_ids",
+        "attribute_line_ids.attribute_id.name",
+    )
+    def _compute_reference_mask(self):
         for rec in self:
-            if automask or rec.reference_mask == "":
+            if self.is_automask() or rec.reference_mask == "":
```
sebastienbeau: call is_automask before the loop, it will avoid to restest the group for each record

sebastienbeau (changes requested):

sebastienbeau:
@pedrobaeza @CasVissers-360ERP let's go for the warning message for this module
Then we will propose an extra module "product_variant_default_code_constraint", that add constraint like (uniq code per attribute, mandatory code, etc...)

pedrobaeza:
Isn't enough with the module for doing unique product code (I think it's called `product_code_unique`)?

Kev-Roche:
@sebastienbeau, I squashed the commits and added your code improvements.

sebastienbeau:
> Isn't enough with the module for doing unique product code (I think it's called `product_code_unique`)?

For some customer it will be enough but for other customer (like the one who fund this work) they want something more constraint
- unique and required code on product.attribute.value per attribute
- unique and required code on product.attribute
- unique and required prefix_code on product.template

I pretty sure that other customer want the same behavior (and also integrator to avoid customer mistake ;) ) so let's share it.
We will propose an other module for that in an other PR

product_variant_default_code/models/product.py
```diff
@@ -100,12 +76,62 @@ class ProductTemplate(models.Model):
...
+        "code_prefix",
+        "attribute_line_ids.value_ids",
+        "attribute_line_ids.value_ids.code",
+        "attribute_line_ids.value_ids.name",
+    )
+    def _compute_variant_default_code_error(self):
+        if self.is_automask():
+            for rec in self:
+                error_txt = ""
+                if not rec.code_prefix and self.is_automask():
```
sebastienbeau: I think my explanation was not clear, I just say that you should init a variable outside the loop to avoid to call it all the time.
error on missing code should be always display even in case of automask

```suggestion
        automask = self.is_automask():
        for rec in self:
            error_txt = ""
            if not rec.code_prefix and automask:
```

sebastienbeau (changes requested):

pedrobaeza:
Great, such module is a good complement indeed.

product_variant_default_code/models/product.py
Kev-Roche: My bad, i did not check all the 'is_automask()' uses

CasVissers-360ERP:
@Kev-Roche 
I have found a different issue, if the reference_mask is updated in the same write as a new default_code is generated the new reference_mask is not used. When changing reference_mask to store = False this doesn't occur anymore. Maybe we should just calculate the reference_mask in _generate_default_code?

Kev-Roche:
@CasVissers-360ERP 
Indeed, I replicate the issue. I was unable to bring up your solution but I found another way. Consequently, if a code_prefix is present, it will appear in the default_code.

sebastienbeau (approved):
LGTM code review and test

PR #228, opened 2021-11-29
[13.0][ADD] product_attribute_value_archive

mmequignon:
Allows to archive a `product.attribute.value` which is referenced by archived `product.product`.

In std odoo, removing a `product.attribute.value` from a `product.template`
either leads to a `product.product` unlink or archive,
depending on if the variant is referenced on a `sale.order.line` or not.
Then, if user tries to remove the `product.attribute.value` from the `product.attribute`,
he will get an error because of the `ondelete restrict` on `product.product`.


product_attribute_value_archive/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,7 @@
+Allows to archive a `product.attribute.value` which is referenced by archived `product.product`.
+
+In std odoo, removing a `product.attribute.value` from a `product.template`
+either leads to a `product.product` unlink or archive,
+depending on if the variant is referenced on a `sale.order.line` or not.
+Then, if user tries to remove the `product.attribute.value` from the `product.attribute`,
+he will get an error because of the `ondelete restrict` on `product.product`.
```
i-vyshnevska, line 7: here you can mention a PR on odoo in case they merge it master to not migrate in future

i-vyshnevska (approved):

product_attribute_value_archive/readme/DESCRIPTION.rst
mmequignon: It will never be merged, I opened this PR so they're aware that this is broken.
The better they can do is to implement it for 16.0 :/

sebalix: High chance they don't do it, perhaps opening a PR on `master` to have better chances...

mmequignon: I'll do that later, no time for this right now.

product_attribute_value_archive/models/product_attribute_value.py
```diff
@@ -0,0 +1,81 @@
...
+from odoo import api, fields, models
+
+
+class ProductAttributeValue(models.Model):
+    _inherit = "product.attribute.value"
+
+    active = fields.Boolean(default=True)
+
+    def _get_related_variants(self):
+        """Returns all related archived products.
```
simahawk, line 13: https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings

product_attribute_value_archive/models/product_template_attribute_value.py
```diff
@@ -0,0 +1,36 @@
...
+        self.write({"ptav_active": True})
+        # get related archived product_attribute_value and unarchive it
+        attr_val = self.product_attribute_value_id
+        if attr_val.active is False:
+            attr_val._unarchive()
+        # Rebuild the product_template_attribute_line link if it has been
+        # archived.
+        tmpl_attr_line = self.attribute_line_id
+        vals = {"value_ids": [(4, attr_val.id, 0)]}
+        if tmpl_attr_line.active is False:
```
simahawk:
```suggestion
        if not tmpl_attr_line.active:
```

product_attribute_value_archive/readme/DESCRIPTION.rst
simahawk: yes pls

simahawk (approved):
LG overall

PR #229, opened 2021-12-10
[13.0][FIX] product_variant_configurator: Drop attribute_id search on creation

HviorForgeFlow:
Steps to reproduce:
- Try to create a product variant manually from Product Variant.

Odoo fails because dict key "attribute_id" has disappeared

Traceback:
image

CC @ForgeFlow

### DIFF
--- a/product_variant_configurator/models/product_template.py
+++ b/product_variant_configurator/models/product_template.py
@@ -98,3 +98,12 @@ class ProductTemplate(models.Model):
                 if limit and len(res) >= limit:
                     break
         return res
+
+    def _get_variant_for_combination(self, combination):
+        """
+        Overwrite method to allow attributes that no create variants be selectable
+        """
+        self.ensure_one()
+        return self.env["product.product"].browse(
+            self._get_variant_id_for_combination(combination)
+        )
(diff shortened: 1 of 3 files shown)


pedrobaeza:
Steps to reproduce the problem?

HviorForgeFlow:
Try to create a product variant manually from Product Variant, dict key "attribute_id" has disappeared

pedrobaeza:
Please include them on the commit message, with full details, for later reference.

product_variant_configurator/models/product_template.py
```diff
@@ -98,3 +98,12 @@ def name_search(self, name="", args=None, operator="ilike", limit=100):
                 if limit and len(res) >= limit:
                     break
         return res
+
+    def _get_variant_for_combination(self, combination):
+        """
+        Overwrite method to allow attributes that no create variants be selectable
+        """
+        self.ensure_one()
+        return self.env["product.product"].browse(
```
pedrobaeza, line 107: This overwriting shouldn't be absolute, but only activated with a flag or something, but calling super in the rest of the cases.

HviorForgeFlow: @pedrobaeza I will take a look later on

PR #233, opened 2021-12-17
[13.0][IMP] product_variant_configurator - edit attributes values con variants

ChrisOForgeFlow:
Added possibility to edit attributes values con variants, with specific permission, and template configured as No Create Templates
@ForgeFlow

Todo:

* [ ] Rename module to product_variant_configurator_editable_attribute


JordiBForgeFlow:
@ChrisOForgeFlow A user not in group "Allow to edit attribute values in variants" can still edit the attribute values in the variants, and in that case the list is not correctly filtered. The field must be read-only if you are not in this group.

image

pedrobaeza:
This is very dangerous and I don't think this should be added in this module.

PR #234, opened 2021-12-29
[14.0] [MIG] purchase_order_variant_mgmt: Migrate to version 14.0

sonhd91:


leemannd (approved):

purchase_order_variant_mgmt/models/purchase_order.py
```diff
@@ -17,8 +17,8 @@ class PurchaseOrderLine(models.Model):
     state_purchase_order_variant_mgmt = fields.Selection(
         related="order_id.state", readonly=True, string="Order status"
     )
-    product_attribute_value_ids = fields.Many2many(
-        comodel_name="product.attribute.value",
-        related="product_id.attribute_value_ids",
+    product_template_attribute_value_ids = fields.Many2many(
```
rousseldenis, line 20: @sonhd91 I'm wondering in which measure the field name shouldn't be kept like before as it will maybe add some work to people during migration if they inherited from this.

IMHO we can keep field name, changing just the comodel_name and related field attributes.

hailangvn: Thanks @rousseldenis for your comment. I think the main driver for the change is consistency. By looking at `product_template_attribute_value_ids` of the model, we can know where it comes from.

pedrobaeza:
Is this still needed with the new product matrix configurator in native core?

hailangvn:
Good day @pedrobaeza, in native core I found only searching feature in purchase order line as in below screenshot. May I know what is the matrix configuration you mentioned?
v14 0_po_search_product_variants

To compare, below is the wizard from this module. That will ease the process of adding multiple variants of a product that has up to 2 attributes.

v14 0_purchase_order_variant_mgmt-wizard

For products 3 or above attributes, the wizard doesn't work.
v14 0_purchase_order_variant_mgmt-wizard-3_attributes
v14 0_product-3_attributes

pedrobaeza:
It's the modules `product_matrix` and `purchase_product_matrix`: https://github.com/odoo/odoo/tree/14.0/addons/purchase_product_matrix

Selección_024

Selección_026

Selección_027

And this one serves for 2 or more attributes.

hailangvn:
Thank you for the great guide. It is really better than this module.

PR #235, opened 2022-01-03
[12.0] IMP product_variant_sale_price handling skip_update_fix_price at create time

eLBati:
Following what is done at write


product_variant_sale_price/models/product_product.py
```diff
@@ -15,6 +15,8 @@ def _update_fix_price(self, vals):
     @api.model
     def create(self, vals):
         product_tmpl = super(ProductTemplate, self).create(vals)
+        if self.env.context.get('skip_update_fix_price', False):
```
rousseldenis, line 18: @eLBati Maybe more readable and coherent in the _update_fix_price function no ?

` if not self.env.context.get('skip_update_fix_price') and list_price' in vals:`

PR #236, opened 2022-01-05
[14.0][MIG] product_attribute_value_archive

mmequignon:


sebalix:
Port of https://github.com/OCA/product-variant/pull/228

product_attribute_value_archive/tests/test_archive_attribute_value.py
```diff
@@ -75,18 +74,22 @@ def setUpClassCommon(cls):
...
+        for product in products:
+            vals.append(
+                {
+                    "product_id": product.id,
+                    "product_uom_qty": 1,
+                    "order_id": cls.order.id,
+                }
+            )
+        cls.env["sale.order.line"].create(vals)
+        return cls.order
```
sebalix, line 92: We should avoid impure functions, here you update a variable (`cls.order`) that is declared outside of the method, while the name of the method and its parameters suggests you are creating order lines only.
Why did you change the previous implementation?

If there is a good reason to change the implementation, I would write:
```python
def _create_sale_order_lines(cls, order, products):
    ...
    return order
```
And pass explicitly the order on each call.

sebalix (changes requested):

product_attribute_value_archive/tests/test_archive_attribute_value.py
mmequignon: The issue is that, sale_stock is installed by another module and introduces a required field on sale_order.
However, we just have the constraint on the database, but the fields doesn't exists in the odoo context.
Then, odoo doesn't set the default value on this fields, and creating sale order fails.
I've found this workaround : adding so lines to on an existing sale order.

product_attribute_value_archive/tests/test_archive_attribute_value.py
```diff
@@ -75,18 +74,22 @@ def setUpClassCommon(cls):
             ]
         )
         cls.red_pen = pen_template_value.ptav_product_variant_ids
+        cls.order = cls.env.ref("sale.sale_order_1")
```
sebalix, line 77: Then this line deserves a comment:
```python
# NOTE: use an existing order and avoid the creation a new one
# to avoid NOT NULL constraint introduced by other addons
```

sebalix (changes requested):

PR #237, opened 2022-01-05
[14.0][MIG] product variant change attribute value

mmequignon:
backport of https://github.com/OCA/product-variant/pull/222

### DIFF
--- a/product_variant_change_attribute_value/README.rst
+++ b/product_variant_change_attribute_value/README.rst
@@ -1,0 +1,1 @@
+wait for the bot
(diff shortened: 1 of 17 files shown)


sebalix:
Note: Git history will need some cleanup on v13 PR before merging this (we will update the current v14 PR accordingly right after).

rousseldenis:
@mmequignon Thanks for this.

Could you just fix commits messages:

image


The migration one should be like PR title ([14.0][MIG] product_variant_change_attribute_value)


By the way, the module version change should normally be in migration commit. But that's ok here.

PR #238, opened 2022-01-12
[14.0] [IMP] product_variant_change_attribute_value: Change module name product_attribute_manager

sonhd91:
* Base on [PR#237](https://github.com/OCA/product-variant/pull/237)
* Add feature remove all attribute

### DIFF
--- a/product_variant_change_attribute_value/README.rst
+++ b/product_variant_change_attribute_value/README.rst
@@ -1,0 +1,1 @@
+wait for the bot
--- a/product_variant_change_attribute_value/__init__.py
+++ b/product_variant_change_attribute_value/__init__.py
@@ -1,0 +1,2 @@
+from . import tests
+from . import wizards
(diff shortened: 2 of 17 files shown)


rousseldenis:
> * Change module name to `product_attribute_manager`

@sonhd91 This is quite a regression as IMHO 'manager' is too vague as so many things can be hidden behind that term.

We can keep the former name even if adding the removal feature (as it can be understood as a 'change').

jgrandguillaume (approved):
Hi,

Functional test ok, I agree the original name would be better thought.

Joël

sonhd91:
> Hi,
> 
> Functional test ok, I agree the original name would be better thought.
> 
> Joël


> Hi,
> 
> Functional test ok, I agree the original name would be better thought.
> 
> Joël

Sorry for the missing out of this PR, branch updated, keeping module name `product_variant_change_attribute_value`

Issue #239, opened 2022-01-12
[BUG][13.0] product_variant_configurator makes it possible to create product.[product|template] without name


Rad0van:
## Module
product_variant_configurator

## Describe the bug
When using product_variant_configurator it is possible to create new records without required fields.

## To Reproduce
**Affected versions**: 13.0

1. Create import file with just 1 columns: internal_reference Fill it with non-existent Product (Variant) codes.
2. Attempt importing the prepared file. It will create records for both Product (Variant) and Product Template without required fields - i.e. name.

**Expected behavior**
This should fail on missing required attribute - name.

**Additional context**
There is some magic I yet do not understand going on which inserts empty name into records being created:
https://github.com/OCA/product-variant/blob/e0da7b28ad5c41735525e3c63313f2929bc9a50e/product_variant_configurator/models/product_template.py#L47-L52
https://github.com/OCA/product-variant/blob/e0da7b28ad5c41735525e3c63313f2929bc9a50e/product_variant_configurator/models/product_product.py#L178

@zakiuu @HviorForgeFlow @rvalyi @sbidoul @chienandalu @AdriaGForgeFlow any ideas how to tackle this? I am planning to work on this as this causes troubles for me but any help would be welcome.

Rad0van:
Solved by #240.

PR #240, opened 2022-01-12
[13.0][FIX] product_variant_configurator: Fix creation of incorrect records when no name is provided

Rad0van:
This should be addressing issue #239

### DIFF
--- a/product_variant_configurator/models/product_product.py
+++ b/product_variant_configurator/models/product_product.py
@@ -175,5 +175,7 @@ class ProductProduct(models.Model):
             )
             vals.pop("product_attribute_ids")
             vals["product_template_attribute_value_ids"] = [(4, x) for x in ptav]
-        obj = self.with_context(product_name=vals.get("name", ""))
+        obj = (
+            self.with_context(product_name=vals.get("name")) if "name" in vals else self
+        )
         return super(ProductProduct, obj).create(vals)


pedrobaeza:
Maybe it's better to not inject such context if the value is not in the dictionary?

Rad0van:
> Maybe it's better to not inject such context if the value is not in the dictionary?

You mean something like
`obj = self.with_context(product_name=vals.get("name")) if "name" in vals else self`
I'm sure there's more elegant way to write this in Python but would it still be more elegant than the proposed version? I mean I can see the check in product.template is `if "product_name" in self.env.context` so it would be nice if actually that was how the condition was constructed (presence or lack there of it in context). Please suggest how to write it in a nice Python-ic way...

pedrobaeza:
Your suggestion is elegant enough.

Rad0van:
> Your suggestion is elegant enough.

Not to me :-) How about do the magic with swapping name inside product.template create method and completely remove the context voodoo? So in product.template create method there would be something like this:
```
@api.model 
def create(self, vals): 
     vals["name"] = vals.get("product_name", False)
     return super(ProductTemplate, self).create(vals) 
```
The line in product.product create method would just set that value instead of injecting it into context.

Just an idea. Please comment.

pedrobaeza:
If you do that, please add docstring explaining why it's done, so this is not removed in the future if the usefulness is not seen.

Rad0van:
> If you do that, please add docstring explaining why it's done, so this is not removed in the future if the usefulness is not seen.

OK. But which would you prefer? What would you consider better?

pedrobaeza:
The last one seems OK.

Rad0van:
> The last one seems OK.

Somehow that didn't work. So I stuck with the former. Seems OK. I'll squash the commits together.

pedrobaeza:
Maybe you can create a test case for reproducing the problem and thus avoid regressions and see the better fix.

Rad0van:
> Maybe you can create a test case for reproducing the problem and thus avoid regressions and see the better fix.

What I meant it didn't work was that with that alternative approach the tests failed even with correct inputs. For me this is good enough even though I didn't get that Python elegance I was hoping for. So if it's good enough for you I have lots of importing updated price lists (that with the bug were creating unnamed product templates and variants previously instead of failing).

pedrobaeza (approved):
Please squash

Rad0van:
> Please squash

It was done at the time. Is the commit message OK or that needs to be more informative as well?

pedrobaeza:
I'm afraid not. Please follow commit message guidelines: https://www.odoo.com/documentation/15.0/es/developer/misc/other/guidelines.html#commit-message-structure

Rad0van:
> I'm afraid not. Please follow commit message guidelines: https://www.odoo.com/documentation/15.0/es/developer/misc/other/guidelines.html#commit-message-structure

Done. Thanx for the pointer.

pedrobaeza (approved):

rousseldenis (approved):
This should be covered by a test to avoid regression

MiquelRForgeFlow:
@Rad0van Shouldn't this PR be forward ported?

Rad0van:
> @Rad0van Shouldn't this PR be forward ported?

Not sure but probably yes.

PR #247, opened 2022-02-08
[MIG] product_variant_specific_tax: Migration to 14.0

ellbristow:
Straight forward migration. Very little to change
One QoL change to rename a field for clarity


ellbristow:
Ended up needing to rewrite the unit test setup. The test is still the same, but the setup needed changing

PR #250, opened 2022-03-14
[13.0][FIX] product_variant_configurator: Error to create variant from template

igallart:
When you create a product template with attributes with option
"Variant creation: Don't create them automatically" and after
you create a product variant with the template previously created,
it fails (with values and without values):

File "/mnt/data/odoo-addons-dir/product_variant_configurator/models/product_product.py", line 160, in create
    for x in vals["product_attribute_ids"]
  File "/mnt/data/odoo-addons-dir/product_variant_configurator/models/product_product.py", line 160, in <listcomp>
    for x in vals["product_attribute_ids"]
KeyError: 'attribute_id'

I think that is not necessary add at the domain the value "attribute_id", an attribute_value is linked only a one attribute_id.

Screenshot 2022-03-14 at 13-27-04 PRUEBA - Odoo
Screenshot 2022-03-14 at 13-31-31 New - Odoo
Screenshot 2022-03-14 at 13-27-42 New - Odoo
Screenshot 2022-03-14 at 13-27-52 New - Odoo

### DIFF
--- a/product_variant_configurator/models/product_product.py
+++ b/product_variant_configurator/models/product_product.py
@@ -153,14 +153,6 @@ class ProductProduct(models.Model):
                             ],
                         ),
                         (
-                            "attribute_id",
-                            "in",
-                            [
-                                x[2]["attribute_id"]
-                                for x in vals["product_attribute_ids"]
-                            ],
-                        ),
-                        (
                             "product_attribute_value_id",
                             "in",
                             [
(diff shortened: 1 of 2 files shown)


pedroortega-punt:
@pedrobaeza The tests are carried out in Runbot, is there a problem merging?

pedrobaeza:
Duplicated of #231

pedroortega-punt:
@pedrobaeza Ok, Thanks.

PR #256, opened 2022-04-04
[14.0][ADD] product_matrix_sticky_header

miguel-S73:
reopen https://github.com/OCA/product-variant/pull/201

Module to fix table header when scrolling.

As a result of header being sticky, when scrolling if the parent element has padding,
It displays list above the header and it doesn't look right.
I tried these solutions:

- Inject css to the parent of the element. Currently, it's not supported.
- Access the dialog widget of the product matrix to inject a custom class.
    The widget is not accessible as it is created within a method.
After trying these without success, I ended up with a solution that while not being very elegant, it works
I'm open to any suggestions 

#### BEFORE  `fixOverflowTbodyStickyHeader`
product_matrix_header_sticky_before_fix

#### AFTER  `fixOverflowTbodyStickyHeader`
product_matrix_header_sticky


hildickethan:
rebased

hildickethan (approved):

miguel-S73:
@pedrobaeza Do you see this module OK?

pedrobaeza:
Let's merge it, although this may deserve a fix on Odoo itself. Don't you think?

/ocabot merge nobump

PR #258, opened 2022-05-06
[13.0][FIX] product_variant_configurator: _build_attributes_domain

jo541:
Fully manage dictionaries in _build_attributes_domain.

### DIFF
--- a/product_variant_configurator/models/product_product.py
+++ b/product_variant_configurator/models/product_product.py
@@ -34,7 +34,13 @@ class ProductProduct(models.Model):
     def _build_attributes_domain(self, product_template, product_attributes):
         domain = []
         cont = 0
+        attributes_ids = []
         if product_template:
+            for attr_line in product_attributes:
+                if isinstance(attr_line, dict):
+                    attributes_ids.append(attr_line.get("attribute_id"))
+                else:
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Can you please specify a bit more?

jo541:
At the base the method managed only partially the possibility to spend product_attributes like a dictionary.

Line 40: An if condition check the possibility of the  product_attributes contain dictionaries or records. 
Line 51: Another loop on product_attributes is managed only like records.

pedrobaeza:
When a dict is passed in the current code? Is that something custom on your side?

jo541:
It's in the current code.

Current code :
The method _build_attributes_domain have a param product_attributes, this param is iterated at the line 39. In this loop at the line 40 a check verify if param contain a list of dict or records. This check lets think that param could be a list of dict. 

But at the line 51 a list comprehension iterate on product_attributes. This list comprehension only work in the case of the param product_attributes are records. Not if the product_attributes is a list of dict.

In the MR:

Move the list comprehension at the line 51 to a new loop for at the line 39 to manage the case of the param is a list of dict or records

pedrobaeza (approved):
Not sure if that's dead code on current module, but better to be coherent.

ThomasBinsfeld (approved):

PR #260, opened 2022-05-24
V14 product configuration and Purchasing Product Configuration 

AbdelrahmanHamdy1996:
add Product configuration and  Purchasing Product configurator with tested and fixed  for V14


pedrobaeza:
Thanks for the contribution.

Please preserve commit history following technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-14.0 and only one module per PR.

If the jump is between several versions, you have to modify the source branch in the main command to accommodate it to this circumstance.

PR #241, opened 2022-01-13
[14.0][ADD] product_variant_default_code_constraint module

Kev-Roche:
Extension of the product_variant_default_code module, adding more constraints and restrictions on the differents default_code parts, according to the discussion here : https://github.com/OCA/product-variant/pull/227

PR #262, opened 2022-06-30
[14.0][add] product variant default code constraint

Kev-Roche:
Wrong git rebase manipulation on my side, ...
this PR, replace this one : https://github.com/OCA/product-variant/pull/241

@sebastienbeau


sebastienbeau:
@Kev-Roche Travis is raising an error because when we install the module the constraint can not be set (as we have empty code)

We should add default value in pre install script to be able to set the constraint

francesco-ooops:
@Kev-Roche can you rebase this one too?

Kev-Roche:
@francesco-ooops also done

PR #263, opened 2022-06-30
[14.0][IMP] product_variant_inactive: Improved management of variants

chafique-delli:
- **Improvement of variants removal:** 
When you remove a variant feature then if the variant product has been used, Odoo will disable it and it will remain visible forever in the variants tab.
The idea is to flag these variants for deletion so that they are no longer seen

- **Improved performance when editing articles with many variants**:
Flag the variants whose combination has been deleted, we will no longer try to delete them when editing variants.

- Do not allow the reactivation of a variant whose combination has been deleted.
- Removal of the dependency on the Stock module
- Tests

### DIFF
--- a/product_variant_inactive/models/product_variant.py
+++ b/product_variant_inactive/models/product_variant.py
@@ -5,13 +5,19 @@ import logging
 
 from lxml import etree
 
-from odoo import api, models
+from odoo import _, api, fields, models
+from odoo.exceptions import UserError
 
 _logger = logging.getLogger(__name__)
 
 
 class ProductProduct(models.Model):
     _inherit = "product.product"
+
+    combination_deleted = fields.Boolean(
+        help="If tick then this combination of variant"
+        " has been deleted and can not be reactivated"
+    )
 
     @api.model
     def fields_view_get(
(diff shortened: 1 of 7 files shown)


chafique-delli:
@sebastienbeau

product_variant_inactive/tests/test_product_variant_inactive.py
```diff
@@ -103,3 +110,34 @@ def test_reactive_template(self):
         self.assertEqual(variants.mapped("active"), [False, False])
         template.write({"active": True})
         self.assertEqual(variants.mapped("active"), [True, True])
+
+    def test_remove_combination(self):
+        with mock.patch.object(
+            type(self.env["product.product"]), "unlink", side_effect=UserError
+        ):
+            self.product_product_4c._unlink_or_archive(check_access=True)
```
sebastienbeau: I would prefer here to not call directly the _unlink_or_archive method, but to remove a attribute.value of the product template

product_variant_inactive/tests/test_product_variant_inactive.py
```diff
@@ -103,3 +110,34 @@ def test_reactive_template(self):
...
+        # check that variant is inactive and combination_deleted is True
+        self.assertFalse(self.product_product_4c.active)
+        self.assertTrue(self.product_product_4c.combination_deleted)
+
+    def test_remove_combination_not_activable(self):
+        self.test_remove_combination()
+        message_error = ""
+        try:
+            self.product_product_4c.active = True
```
sebastienbeau, line 126: Do not use try except in test but use with assertRaise  like here https://github.com/OCA/product-variant/blob/a461cb0cdfbb5ef127f693cbeb574bab1e065daf/product_variant_default_code/tests/test_variant_default_code.py#L192

sebastienbeau (changes requested):

product_variant_inactive/models/product_variant.py
```diff
@@ -44,8 +50,40 @@ def fields_view_get(
...
+                super(ProductProduct, product).write(vals)
+        return True
+
+    def _unlink_or_archive(self, check_access=True):
+        records = self
+        if not self._context.get("skip_filter_deleted"):
+            records = records.filtered(
+                lambda p: not p.combination_deleted
+            ).with_context(skip_filter_deleted=True)
+            records.write({"combination_deleted": True})
```
sebastienbeau: You should write "active: False" here it will avoid having an hack in the constraint
```suggestion
            records.write({"combination_deleted": True, "active": False})
```

sebastienbeau (changes requested):

product_variant_inactive/models/product_variant.py
```diff
@@ -44,8 +50,40 @@ def fields_view_get(
...
+        super(ProductProduct, records)._unlink_or_archive(check_access=check_access)
+
+    @api.constrains("active", "combination_deleted")
+    def _check_can_not_be_reativated(self):
+        for record in self:
+            if (
+                not self._context.get("skip_filter_deleted")
```
sebastienbeau: no need of checking the context (see before)

chafique-delli:
@sebastienbeau , all requested changes have been made.

sebastienbeau (approved):
LGTM (code review) and test

Kev-Roche (approved):
Code & fonctional review LGTM

PR #274, opened 2022-10-26
[FIX][product_variant_default_code] check code of active value_ids

Kev-Roche:
Insure that default_code are computed on active value_ids.


francesco-ooops:
@Kev-Roche could you rebase?

pedrobaeza (changes requested):
You can't modify dependencies in a stable module. Create a glue module or find another way.

PR #280, opened 2023-01-28
[FIX][14.0]product_variant_sale_price bad uom computationx

### LINKED ISSUE #259: [13.0][14.0] Wrong price on SO when using UOM with ratio
## Module
product_variant_sale_price

## Describe the bug
Price is not correct on SO on a product configured with a different UOM than the base UOM and we change the UOM to an other UOM of same type on SO

## To Reproduce
**Affected versions**: 13.0, 14.0 (not tested on other versions)

1. Install modules: sale, product_variant_sale_price
2. Create a new UOM "box of 2000 units" with :
             - Category: "Unit"
             - Type: "Bigger than the reference Unit of Measure"
             - Bigger Ratio: "2000"
image

3. Create a product with : 
             - Sales price: 2000000
             - Unit of measure: Box of 2000
image

4.  Create a SO and set product in a sale order line
image

Price is correct
6. Change UOM to "Unit"
image

Price is wrong

**Expected behavior**
Price should have been "2000000 euro / 2000 (ratio)"=1000 like on standard Odoo

**Additional context**
Error seems to be in functions "_compute_lst_price", "_compute_list_price", "_inverse_product_lst_price" on product.product
If you open the product from the sale orrder line with "unit" UOM, you'll see "Public price" already computer with the UOM set on sale order line 
image

flotho:
Force  the use of uom while computing price
FIX #259

### DIFF
--- a/product_variant_sale_price/models/product_product.py
+++ b/product_variant_sale_price/models/product_product.py
@@ -34,20 +34,18 @@ class ProductProduct(models.Model):
         uom_model = self.env["uom.uom"]
         for product in self:
             price = product.fix_price or product.list_price
-            if "uom" in self.env.context:
-                price = product.uom_id._compute_price(
-                    price, uom_model.browse(self.env.context["uom"])
-                )
+            if self.env.context.get("uom"):
+                context_uom = uom_model.browse(self.env.context.get("uom"))
+                price = product.uom_id._compute_price(price, context_uom)
             product.lst_price = price
 
     def _compute_list_price(self):
         uom_model = self.env["uom.uom"]
         for product in self:
             price = product.fix_price or product.product_tmpl_id.list_price
-            if "uom" in self.env.context:
-                price = product.uom_id._compute_price(
-                    price, uom_model.browse(self.env.context["uom"])
-                )
+            if self.env.context.get("uom"):
+                context_uom = uom_model.browse(self.env.context.get("uom"))
+                price = context_uom._compute_price(price, context_uom)
             product.list_price = price
 
     def _inverse_product_lst_price(self):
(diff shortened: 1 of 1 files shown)


flotho:
ping @jlaloux @olimalt85

product_variant_sale_price/models/product_product.py
```diff
@@ -34,20 +34,18 @@ def _compute_lst_price(self):
...
         uom_model = self.env["uom.uom"]
         for product in self:
             price = product.fix_price or product.product_tmpl_id.list_price
-            if "uom" in self.env.context:
-                price = product.uom_id._compute_price(
-                    price, uom_model.browse(self.env.context["uom"])
-                )
+            if self.env.context.get("uom"):
+                context_uom = uom_model.browse(self.env.context.get("uom"))
+                price = context_uom._compute_price(price, context_uom)
```
RabbitJon-S73:
```suggestion
                price = product.uom_id._compute_price(price, context_uom)
```
It should be like above?

flotho:
ping @jlaloux @Olimalt85 @RabbitJon-S73

product_variant_sale_price/models/product_product.py
```diff
@@ -34,20 +34,18 @@ def _compute_lst_price(self):
         uom_model = self.env["uom.uom"]
         for product in self:
             price = product.fix_price or product.list_price
-            if "uom" in self.env.context:
-                price = product.uom_id._compute_price(
-                    price, uom_model.browse(self.env.context["uom"])
-                )
+            if self.env.context.get("uom"):
+                context_uom = uom_model.browse(self.env.context.get("uom"))
```
BT-anieto, line 38: **[suggestion]** Since you already checked that "uom" key is in the context and has a value, you can access direct the dictionary like self.env.context["uom"] instead of using get, since it is faster.

BT-anieto (commented):
Hello! :) I'm just wondering why this fix solves the bug. Could you explain, please?

As far as I see, the only real difference is in the **if** line, since it avoids the computation if _uom_ is in the context, but not set. Is this what it is happening when the unit is changed back to _units_?

PR #283, opened 2023-02-05
product_variant_inactive: fix unlink variant

sebastienbeau:
When deleting a variant with other inactive variant, odoo is going to delete everything.
This can lead to unwanted deletation when using this module, so skip this behaviour, and only delete if there is really no existing variant.


sebastienbeau:
@Kev-Roche

Kev-Roche (approved):

PR #281, opened 2023-02-01
[16.0][MIG] product_variant_sale_price

RabbitJon-S73:
Standard migration to 16.0.

Added https://github.com/OCA/product-variant/pull/280

PR #290, opened 2023-04-13
[16.0][MIG] product_variant_sale_price: Migration to 16.0

ecino:
Adapted the views and fixed a bug where the sales price on the template was updating with the previous value of a product.

This PR includes also the work done by RabbitJon-S73 in #281 but goes a little bit further by improving the views.


ecino:
This PR includes also the work done by RabbitJon-S73 in #281 but goes a little bit further by improving the views.

RabbitJon-S73:
@ecino LGTM, ask the other reviewers to have a look and if they find it ok im closing my PR to avoid conflicts.

hitrosol (approved):
Tested in runboat, LGTM.

HugoCordobaLeal:
Hi! In my opinion, you should propose the improvements in the other PR and have them added in the PR that was done in the first place.

Regards!

enriquemartin (approved):

HekkiMelody:
@ecino Could you rebase and solve the conflicts please?

HekkiMelody:
Warning during installation: 

```
2023-05-11 14:43:45,436 20 WARNING devel odoo.api.create: The model odoo.addons.product_variant_sale_price.models.product_product is not overriding the create method in batch 
```

Could you overwrite the method using the `@api.model_create_multi` decorator? It shouldn't be too much work, and in fact it will make create and write more similar as a result

ecino:
Thanks for the reviews and sorry for the delay in my answer. I corrected all reported issues.

HekkiMelody (approved):
Code review, LGTM

product_variant_sale_price/models/product_product.py
```diff
@@ -0,0 +1,81 @@
...
+            product.write(vals)
+
+    lst_price = fields.Float(
+        compute="_compute_lst_price",
+        inverse="_inverse_product_lst_price",
+    )
+    list_price = fields.Float(
+        compute="_compute_list_price",
+    )
+    fix_price = fields.Float()
```
FrankC013: ### Proposed Changes for Implementation

These proposals are aimed at how this new implementation will impact the _price_extra_ field of the **product** module. 
The _price_extra_ field, with the changes you have proposed, will stop being displayed but will not erase its data, a fact with which I agree, since if the module is uninstalled, access should still be had to the previously stored information. However, the **sale.order.line** module calculates the _price_unit_ by adding the value of _price_extra_ and this can generate inconsistencies if the field has old data stored.

https://github.com/odoo/odoo/blob/5ac58ebf983c1c02019253c0430fde3502e13e7c/addons/product/models/product_product.py#L609-L628

To avoid each user having to perform a data migration, I think the most appropriate would be to modify the __compute_product_price_extra()_ method of the _price_extra_ in the **product.product** to ensure that in other models it does not take the obsolete value from the database. The **product.product**'s _price_extra_ field is a `store=False`, so it will not impact the information of the _price_extra_ stored in the **product.template.attribute.value**. 
For this reason, I propose add the following change:

```python
    def _compute_product_price_extra(self):
        for product in self:
            product.price_extra = 0.0
```

product_variant_sale_price/models/product_product.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def write(self, vals):
+        res = super().write(vals)
+        if self.env.context.get("skip_update_fix_price", False):
+            return res
+        for template in self:
+            template._update_fix_price(vals)
+        return res
+
+
```
FrankC013, line 50: On the other hand, within the sale module it would be advisable to hide the badge that is generated when the _price_extra_ is informed.

https://github.com/odoo/odoo/blob/5ac58ebf983c1c02019253c0430fde3502e13e7c/addons/sale/views/variant_templates.xml#L119C1-L142

For this reason and to continue with the structure of our changes, I propose modifying the __get_combination_info()_ method so that it always reports that the `price_extra = 0.0`

```suggestion
    def _get_combination_info(self, combination=False, product_id=False, add_qty=1, pricelist=False, parent_combination=False, only_template=False):
        res = super(ProductTemplate, self)._get_combination_info(combination, product_id, add_qty, pricelist, parent_combination, only_template)
        res['price_extra'] = 0.0
        return res
```

I await your feedback to know what you think about this!! :D

FrankC013 (changes requested):

product_variant_sale_price/models/product_product.py
eantones: LGTM!

product_variant_sale_price/models/product_product.py
eantones: LGTM!

ecino: Thank you for the very good suggestions. I applied the suggested changes.

ecino:
I applied the nice suggestions from @FrankC013 and also added a new Action that can be very handy for controlling which variant will be the price reference in case the user prefers not to have the cheapest variant as the reference price (which is the default behaviour).

ecino:
Is there something else preventing the merge? Can someone give the final review here? Thanks!

HekkiMelody (approved):

maxgrassi:
Bug report: Tested and this is not working in version 16: "This module also hides sale price at product template level when has more than one variant."
The product price can be set at product templete level and this overwrites all the prices that have been set at product variant level.

maxgrassi:
Suggestion: "Compare to Price?" should be moved at variant level.

ecino:
> Bug report: Tested and this is not working in version 16: "This module also hides sale price at product template level when has more than one variant." The product price can be set at product templete level and this overwrites all the prices that have been set at product variant level.

Where did you quote this from? In my perspective it's not a bug but the way it was made. A tooltip is also displayed for warning about this behaviour, which is totally fine in my opinion.

ecino:
> Suggestion: "Compare to Price?" should be moved at variant level.

This is a good suggestion, however the "Compare to price" field is defined in the "Website sale" module which would add a potential not needed dependency. A new bridge module extending this functionality would be better than integrating this inside this module.

HekkiMelody (approved):

maxgrassi:
> > Bug report: Tested and this is not working in version 16: "This module also hides sale price at product template level when has more than one variant." The product price can be set at product templete level and this overwrites all the prices that have been set at product variant level.
> 
> Where did you quote this from? In my perspective it's not a bug but the way it was made. A tooltip is also displayed for warning about this behaviour, which is totally fine in my opinion.

I Took it from here: https://github.com/ecino/product-variant/tree/14.0/product_variant_sale_price 
I understand this is V14 and in V16 is going to be different however it is much safer to "avoid to edit" the sale price at product template level. If you have complex variants with many variants and many different prices and you touch by mistake the price in the product template (can happen), you end to loose all the prices set in the variants. This is a very dangerous behavior in everyday use. At least leave the possibility to turn off the  edit mode for the price in the product template level.

eantones:
I think the change https://github.com/OCA/product-variant/commit/39c984ced2376a37b435b866925822e45e8499c3#diff-9928710829783336eb5c9e25e03423f41edc07f81ac7051fbb6e799e8a2394f6 you introduced after the fixes proposed by @FrankC013 is getting in the way for me to approve this PR. It introduces a new logic that needs to be understood and tested, thereby delaying the overall PR approval. But, anyway, in my opinion, this change is indeed an improvement and it should be in a separate PR, as the current one is solely for the module's migration. If that particular commit were absent, my approval would be immediate.
In any case, thanks @ecino for your job :-)

ecino:
> Suggestion: "Compare to Price?" should be moved at variant level.

@maxgrassi Here is your feature request, in a separate module : #297

ecino:
> > > Bug report: Tested and this is not working in version 16: "This module also hides sale price at product template level when has more than one variant." The product price can be set at product templete level and this overwrites all the prices that have been set at product variant level.
> > 
> > 
> > Where did you quote this from? In my perspective it's not a bug but the way it was made. A tooltip is also displayed for warning about this behaviour, which is totally fine in my opinion.
> 
> I Took it from here: https://github.com/ecino/product-variant/tree/14.0/product_variant_sale_price I understand this is V14 and in V16 is going to be different however it is much safer to "avoid to edit" the sale price at product template level. If you have complex variants with many variants and many different prices and you touch by mistake the price in the product template (can happen), you end to loose all the prices set in the variants. This is a very dangerous behavior in everyday use. At least leave the possibility to turn off the edit mode for the price in the product template level.

@maxgrassi I think it's best to let the user decide. A warning message is displayed below the field and is enough in my opinion.

image

ecino:
@eantones @FrankC013 Separated the features additions in #298 

This PR is only the migration of the module.

eantones (approved):

eantones (changes requested):
@ecino the code added by @FrankC013 were not new features, they were fixes so they should be included in the migration module

ecino:
Sorry, I misunderstood your last comment then. Now I think I get it, you only want to split the server action I added, right?

eantones:
Yes, thanks

eantones:
LGTM

eantones (approved):

FrankC013 (approved):

HekkiMelody:
@OCA/product-maintainers can this be merged please?

willmakenoise:
@OCA/product-maintainers Any chance this can be merged? 😀

eantones:
@OCA/product-maintainers can we merge this PR? :-)

hildickethan (approved):

hildickethan:
It would be better if the last commit followed the usual commit guidelines
`[IMP] product_variant_sale_price: Set price_extra to 0` instead of `[IMP] Apply suggested changes from FrankC013 to put price_extra to 0.0`
You can add who suggested the changes in the description of the commit rather than the message itself

manuelcalerosolis:
LGTM

manuelcalerosolis (approved):

fuentes73:
@ecino could you resolve the conflicts? You don't need to edit the README.md it's done by the oca bot.

ecino:
PR Fixed.

LoisRForgeFlow:
@ecino It seems that tests are failing, can you have a look?

ecino:
Apparently it causes problem with the `product_variant_name` module that was introduced lately. I will check when I have time.

uvid-sinisa:
@ecino Looks like module product_variant_name adds field "Name" to product.product as required but tests in product_variant_sale_price create product_products without name.
Would the appropriate fix be to just add some name when creating the product_product in tests?

ecino:
Thanks for the hint. Can we merge now?

HekkiMelody:
@OCA/product-maintainers

HaraldPanten:
@etobella could you merge this one? I forgot I don't have permissions here! 😂

HaraldPanten (approved):

PR #287, opened 2023-03-13
[MIG] product_variant_default_code: migration to 16.0

Rferri44-S73:
Standard migration to 16.0

Depends https://github.com/OCA/product-attribute/pull/1301

PR #292, opened 2023-05-03
[16.0] Revert "[FIX][product_variant_default_code] check code of active value_ids"

pedrobaeza:
This reverts commit 20320aa0abc6ad7ef347d5e2a4b52010e45c4513.

This wild dependency has been introduced silently while the PR proposing it is not accepted and it's not a sane way to keep Odoo modularity. It this should be handled, it can be done putting a hook and a glue module, or searching for other solution.

@Tecnativa


pedrobaeza:
As stated in https://github.com/OCA/product-variant/pull/287#issuecomment-1533320627, reverting this:

/ocabot merge patch

PR #293, opened 2023-05-04
[FIX] product_variant_default_code: code_prefix in reference_mask

RabbitJon-S73:
When `Reference Mask` is edited or `Code prefix` is erased, then the `Code prefix` gets duplicated in `Reference mask`. It can be tedious to work with so I add a condition that check if `Code prefix` is already on `Reference Mask`.

https://user-images.githubusercontent.com/100672471/236146919-c9ec9a12-4e1d-4c34-b443-3ac43d4c23f7.mp4


Kev-Roche (approved):
LGTM

ioans73 (approved):

RabbitJon-S73:
@Kev-Roche hello, is it possible to merge?

PR #291, opened 2023-05-03
[15.0][MIG] product_variant_default_code: Migration to version 15.0

pilarvargas-tecnativa:
cc @Tecnativa TT36978

@stefan-tecnativa @chienandalu please review

PR #294, opened 2023-05-05
[14.0][FIX] product_variant_default_code: fix template error

pilarvargas-tecnativa:
When a product is being created or edited and attributes have not yet been added, a message is printed out in the band in case the reference prefix has not been filled in. If there is no error, this strip should remain hidden.

This error in the template has been detected and fixed in a migration to v15 (PR: #291) and the same problem is being fixed in v14. 

image

cc @Tecnativa TT42976

@stefan-tecnativa @chienandalu please review :)

### DIFF
--- a/product_variant_default_code/models/product.py
+++ b/product_variant_default_code/models/product.py
@@ -109,7 +109,7 @@ class ProductTemplate(models.Model):
                 )
             if error_txt:
                 error_txt = "Default Code can not be computed.\n" + error_txt
-            rec.variant_default_code_error = error_txt
+            rec.variant_default_code_error = error_txt or False
 
     @api.depends(
         "code_prefix",


Kev-Roche (approved):
LGTM

pedrobaeza (approved):
/ocabot merge patch

pedrobaeza:
I think this should be put also in 16.0 and in the 15.0 PR

pilarvargas-tecnativa:
> I think this should be put also in 16.0 and in the 15.0 PR

you mean the commit? in v16 I incorporated it from my local, in v15 I made the change in the migration but the commit is not there, shall I incorporate it?

pedrobaeza:
Everything done, thanks.

PR #295, opened 2023-05-06
[16.0][FIX] product_variant_default_code: fix template error

pilarvargas-tecnativa:
When a product is being created or edited and attributes have not yet been added, a message is printed out in the band in case the reference prefix has not been filled in. If there is no error, this strip should remain hidden.

This error in the template has been detected and fixed in a migration to v15 (PR: https://github.com/OCA/product-variant/pull/291) and the same problem is being fixed in v14 and v16.

image

cc @Tecnativa TT42976

@stefan-tecnativa  @chienandalu  please review :)


pedrobaeza (approved):
Fast-tracking as straight-forward and already merged in previous branches.

/ocabot merge patch

PR #309, opened 2023-10-26
[ADD] product_internal_reference_generator

ilyasProgrammer:
This module allows to generate internal references for Product templates and variants using sequences, setting codes as read-only.

In product template, it's possible to choose among different Internal Reference Templates related to a sequence, and then generate an internal reference with the following structure:

Internal Reference Prefix + progressive number for variant, eg: “Main0001001”, where:

"Main0001" is the prefix generated by sequence and assigned to product template, and

"001" the variant identifier.

Every time a new variant is created, a new internal reference is automatically assigned with progressive variant code.

A specific access rights allows specific users to change internal reference template for a product template once an internal reference has been generated, as well as editing existing ones.


francesco-ooops (approved):
Functional ok!

product_internal_reference_generator/__manifest__.py
```diff
@@ -0,0 +1,18 @@
+{
+    "name": "Product Internal Reference Generator",
+    "summary": """Product template and variant reference based on sequence""",
+    "author": "Ilyas, Ooops, Odoo Community Association (OCA)",
+    "license": "LGPL-3",
```
HekkiMelody: **chore**: License should be AGPL

product_internal_reference_generator/__manifest__.py
```diff
@@ -0,0 +1,18 @@
...
+    "website": "https://github.com/OCA/product-variant",
+    "category": "Sale",
+    "version": "14.0.1.0.0",
+    "depends": ["stock"],
+    "data": [
+        "security/groups.xml",
+        "security/ir.model.access.csv",
+        "views/product.xml",
+    ],
+    "demo": ["demo/demo.xml"],
```
HekkiMelody: **chore**: Please give files meaningful names according to the guidelines

https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#file-naming

product_internal_reference_generator/models/product_template.py
```diff
@@ -0,0 +1,54 @@
...
+    default_code_editable = fields.Boolean(compute="_compute_default_code_editable")
+
+    def _compute_default_code_editable(self):
+        for rec in self:
+            if self.env.user.has_group(
+                "product_internal_reference_generator.group_int_ref_template_always_visible"
+            ):
+                rec.default_code_editable = True
+            else:
+                rec.default_code_editable = not bool(rec.default_code)
```
HekkiMelody: **suggestion**: decorate with `@api.depends_context('uid')`

**suggestion(non-blocking)**: couldn't this have also been achieved by having the field twice with
`groups="product_internal_reference_generator.group_int_ref_template_always_visible"` and
`groups="!product_internal_reference_generator.group_int_ref_template_always_visible"` ?

product_internal_reference_generator/tests/test_product_internal_reference_generator.py
```diff
@@ -0,0 +1,70 @@
...
+        ]
+        df = str(self.pt_seq.sequence_id.number_next_actual - 1) + "003"
+        self.assertIn(df, self.pt_car.product_variant_ids[0].default_code)
+        df = str(self.pt_seq.sequence_id.number_next_actual - 1) + "004"
+        self.assertIn(df, self.pt_car.product_variant_ids[1].default_code)
+        df = str(self.pt_seq.sequence_id.number_next_actual - 1) + "005"
+        self.assertIn(df, self.pt_car.product_variant_ids[2].default_code)
+        df = str(self.pt_seq.sequence_id.number_next_actual - 1) + "006"
+        self.assertIn(df, self.pt_car.product_variant_ids[3].default_code)
+        self.assertFalse(self.pt_plane.get_variant_next_default_code())
```
HekkiMelody: **suggestion(non-blocking)**: consider adding a test that uses the Form class (also so you don't have to call onchanges manually)

product_internal_reference_generator/views/product.xml
```diff
@@ -0,0 +1,129 @@
...
+                    <field name="int_ref_template_id" />
+                    <button
+                        name="btn_generate_sequence"
+                        type="object"
+                        string="Generate"
+                        attrs="{'invisible': [('int_ref_template_id', '=', False)]}"
+                        confirm="Are you sure want to generate new reference prefix?
+                        If variants already exist for this product, the new internal
+                        reference will be assigned only those where internal
+                        reference is not set."
```
HekkiMelody: **polish**

```suggestion
                        confirm="Are you sure want to generate a new reference prefix?
                        If variants already exist for this product, the new internal
                        reference will be assigned only to those variants that
                        do not already have an internal reference set."
```

product_internal_reference_generator/__manifest__.py
```diff
@@ -0,0 +1,18 @@
+{
+    "name": "Product Internal Reference Generator",
```
HekkiMelody, line 2: **question**: should this `exclude` `product_variant_default_code`?

product_internal_reference_generator/views/product.xml
```diff
@@ -0,0 +1,129 @@
...
+                    attrs="{'invisible': [('int_ref_template_id', '=', False)]}"
+                    string="Int. Ref. Prefix"
+                />
+                <field name="default_code_editable" invisible="1" />
+            </field>
+            <field name="default_code" position="attributes">
+                <attribute name="attrs">{
+                    'readonly': [('default_code_editable', '!=', True)],
+                    'invisible': [('product_variant_count', '>', 1)]
+                }</attribute>
```
HekkiMelody: **suggestion**: consider adding a dependency to https://github.com/OCA/server-tools/blob/14.0/base_view_inheritance_extension so you can only add the `readonly` key and not touch the `invisible` key. This would improve the compatibility of this module with other modules that also change the `attrs` of`default_code`

HekkiMelody: Also in two other instances further down.

HekkiMelody (changes requested):
Code review, a few suggestions.

Thanks for your contribution, this must not have been easy to implement!

product_internal_reference_generator/__manifest__.py
ilyasProgrammer: Thanks for review! Yes these modules makes no sence to be intalled together.

product_internal_reference_generator/models/product_product.py
```diff
@@ -0,0 +1,17 @@
...
+    _inherit = "product.product"
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        for vals in vals_list:
+            if vals.get("product_tmpl_id"):
+                pt = self.env["product.template"].browse(vals["product_tmpl_id"])
+                if pt.variants_sequence_id:
+                    vals["default_code"] = pt.get_variant_next_default_code()
+        res = super().create(vals_list)
```
geomer198: Could you remove variable `res`. Please do like this:
`return super(ProductProduct, self).super(vals_list)`

product_internal_reference_generator/models/product_template.py
```diff
@@ -0,0 +1,44 @@
...
+    def btn_generate_sequence(self):
+        self.ensure_one()
+        int_ref_next_val = self.int_ref_template_id.sequence_id.next_by_id()
+        self.variants_prefix = int_ref_next_val
+        var_seq = self.env["ir.sequence"].create(
+            {
+                "name": "variants " + int_ref_next_val,
+                "padding": self.int_ref_template_id.variant_reference_numbers,
+            }
+        )
```
geomer198, line 29: Could you use `write` function:
```py
self.write({
    "variants_prefix": int_ref_next_val,
    "variants_sequence_id": var_seq.id,
    "default_code": int_ref_next_val + var_seq.get_next_char(0)
})

File: product_internal_reference_generator/models/product_template.py [hunk 10]
```diff
@@ -0,0 +1,44 @@
...
+        self.variants_sequence_id = var_seq
+        self.default_code = int_ref_next_val + var_seq.get_next_char(0)
+        self.update_variants_default_code()
+
+    def update_variants_default_code(self):
+        for pp in self.product_variant_ids:
```
geomer198: I think better solution like this:
```py
for pp in self.product_variant_ids.filtered(lambda p: not p.default_code):
    pp.default_code = self.get_variant_next_default_code()
```

product_internal_reference_generator/models/product_template.py
```diff
@@ -0,0 +1,44 @@
...
+            if not pp.default_code:
+                pp.default_code = self.get_variant_next_default_code()
+
+    def get_variant_next_default_code(self):
+        if self.variants_sequence_id:
+            res = self.variants_prefix + self.variants_sequence_id.next_by_id()
```
geomer198: Could you remove `res` variable?
`return self.variants_prefix + self.variants_sequence_id.next_by_id()`

product_internal_reference_generator/tests/test_product_internal_reference_generator.py
```diff
@@ -0,0 +1,63 @@
+# Copyright 2023 Ooops - Ilyas
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo.tests import Form, SavepointCase
+
+
+class TestProductInternalReferenceGenerator(SavepointCase):
```
geomer198, line 6: Could you add docstring for each test cases and message for each assert?

geomer198 (commented):
Please fix my comments.

HekkiMelody (approved):
Code review, LGTM. Nice refactor in the tests!

geomer198 (approved):
LGTM!

francesco-ooops (approved):
Functional is good!

francesco-ooops:
@pedrobaeza can we proceed with this merge? thank you!

pedrobaeza:
What's the difference with https://github.com/OCA/product-attribute/tree/14.0/product_sequence and why here instead of OCA/product-attribute?

francesco-ooops:
Hi @pedrobaeza ,

1) first of all, we published here since it has a similar goal as https://github.com/OCA/product-variant/tree/14.0/product_variant_default_code so we thought this was the most correct repo - It's not easy to understand the difference between the two repos as both miss a repo description... 
We wouldn't mind switching repo if requested.

2) secondly, yes in this case I completely missed the existence of the module you pointed out! 
But once again it's difficult from a functional perspective to evaluate whether an existing module does what you need when documentation only tells you how to setup, but not what the module actually does or the use cases that are covered - as is the case for the module you linked. 

I tested product_sequence and it has a similar approach, but with the following downsides compared to this module:

- 	Sequence is only managed by product category, you cannot have two products in the same category with a different sequence (on the contrary, this module can be extended to automatically assign a reference prefix - and we’ll probably do that)

- 	Once an internal reference has been assigned to a product template, I don’t see any established flow to change it, as an error is raised on product.product . So if you assign the wrong category to the product by mistake, you’re stuck with the wrong sequence

- Internal reference is not read-only for most users thus allowing getting out of the sequence, also allowing the risk of assigning a code that has yet to be assigned by the sequence - leading to an error further down the road

I agree that overall they are not very different, but also they're not the same. I think this could be merged and then see if there will be interest in future migrations to keep only one or both.

What do you think? :)

pedrobaeza:
OK, as asking you to integrate the missing features into the other module, you should at least move it to product-attribute repo. This one is for features specific to handle product variant things. The other module you pointed is because the references are assigned to the variants. It's true that the `-attribute` part of the repo name confuses, but it was done with other purpose, as the original idea was to have a repo for the product and its attributes, but not the variant attributes. For example, the manufacturer of the product, and similar things.

francesco-ooops:
@pedrobaeza thanks for your feedback, but it's not clear to me what's the next step: should we move this PR in product-attribute? 

Is it ok to keep this as a new module? I don't think it's worth the effort try and improve the old one.

ilyasProgrammer:
moved to https://github.com/OCA/product-attribute/pull/1453

PR #316, opened 2023-10-31
[FIX] Conditioning Unique variant and Current selection values use FIX #315

### LINKED ISSUE #315: Opening the Product Variant creation menu from the Product Variant view throws error
An out of range index error prevents you from creating Variants from the Product Variant Menu.

## Module
product_variant_configurator

## Describe the bug
The application prevents the menu to appear due to an out of range index error related to product_variant_ids. Is the Product Configurator model treating the menu as a product template record without attributes?

## To Reproduce
Current version, locally or runboat.

1. Install product_variant_configurator locally
2. Go to the Product Variants menu (You can go from Inventory, Sales, Manufactury..., not relevant)
3. Click on "Create product variant" button
4. You'll trigger the following error, preventing you to open the creation menu.
image

**Expected behavior**
It's supposed to show the Product Variant creation menu.

**Additional context**
I've found the bug both locally and on runboat

joelkm:
Product Variant Configurator seems to treat the Product Variant Creation Menu as a product template record, so its data goes through the execution flow as a supposed product template without attributes (tries to use the unique variant and tries to restrict product possible values to a non-existing selection).

This contribution prevents from using unique variants and values from the selection if they don't exist. It fixes #315  "Product Variant Creation menu not showing" issue without breaking the expected functionality.

### DIFF
--- a/product_variant_configurator/models/product_configurator.py
+++ b/product_variant_configurator/models/product_configurator.py
@@ -97,8 +97,9 @@ class ProductConfigurator(models.AbstractModel):
             return {"domain": {"product_id": []}}
 
         if not self.product_tmpl_id.attribute_line_ids:
-            # template without attribute, use the unique variant
-            self.product_id = self.product_tmpl_id.product_variant_ids[0].id
+            # template without attribute, use the unique variant if exists (prevents from crashing on variant creation in the product variants menu)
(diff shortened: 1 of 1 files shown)


giarve:
@ChrisOForgeFlow you might be interested in this one

PR #325, opened 2023-12-27
[16.0][FIX] product_variant_configurator: implement model_create_multi

alexbottemanne:


product_variant_configurator/models/product_attribute_value.py
```diff
@@ -7,19 +7,20 @@
...
-    @api.model
-    def create(self, vals):
+    @api.model_create_multi
+    def create(self, vals_list):
         """Link created attribute value to the associated template if proceed.
 
         This happens when quick-creating values from the product configurator.
         """
-        attr_value = super(ProductAttributeValue, self).create(vals)
+        attr_value = super(ProductAttributeValue, self).create(vals_list)
```
pedrobaeza: Be consistent in the variable names:

```suggestion
        attr_values = super(ProductAttributeValue, self).create(vals_list)
```

product_variant_configurator/models/product_configurator.py
```diff
@@ -215,25 +215,26 @@ def _get_product_description(self, template, product, product_attributes):
...
-                gen_dict = {
-                    "owner_model": self._name,
-                    "product_tmpl_id": product.product_tmpl_id.id,
-                }
-                for att_val in product._get_product_attributes_values_dict():
-                    att_val.update(gen_dict)
-                    vals["product_attribute_ids"].append((0, 0, att_val))
-        return super().create(vals)
+        for vals in vals_list:
+            if vals.get("product_id"):
```
pedrobaeza: For avoiding a big diff, do this (the same for following occurrences):

```suggestion
            if not vals.get("product_id"):
                continue
```

product_variant_configurator/models/product_template.py
```diff
@@ -44,13 +44,6 @@ def onchange_no_create_variants(self):
                 }
             }
 
-    @api.model
-    def create(self, vals):
```
pedrobaeza, line 48: Why are you removing this?

alexbottemanne: It seemed deprecated; we supply product_name in the context within the create method in product_product, but how is this supposed to influence the name in product_template? I could not find mention of it elsewhere, but I can look into it.

pedrobaeza: Then don't remove the test that is checking this, and specify in the docstring of the test that is for ensuring the correct behavior, which was not correct in the past.

pedrobaeza: And please do this in other commit, as it's something unrelated.

alexbottemanne: I've made another commit with the changes to the restored unit test (or did you mean another pull request?).

product_variant_configurator/tests/test_product_variant_configurator.py
```diff
@@ -595,10 +595,13 @@ def test_product_template_write(self):
...
 
     def test_product_template_create(self):
+        # In the past, the product_template's name would
+        # be incorrectly set to the value of product_name
+        # if present in the context
         product = self.product_template.with_context(
             product_name="Context product name"
         ).create({"name": "Test"})
-        self.assertEqual(product.name, "Context product name")
+        self.assertEqual(product.name, "Test")
```
pedrobaeza: This proves that removing it it's not correct, as the test have changed. You should keep it.

alexbottemanne: I've restored the create method on product_template, in that case, and simply implemented the model_create_multi decorator as with the other creates.

pedrobaeza (approved):
Thanks

giarve (approved):

product_variant_configurator/models/product_template.py
```diff
@@ -44,12 +44,13 @@ def onchange_no_create_variants(self):
...
+    @api.model_create_multi
+    def create(self, vals_list):
         if "product_name" in self.env.context:
```
JordiMForgeFlow, line 49: Hi @pedrobaeza and @alexbottemanne 

Where is this context set in the logic?

pedrobaeza: AFAIK, it was a mechanism for assuring the proper name when creating the template from the variant itself. Not sure if it has been perverted through versions or not needed anymore.

PR #328, opened 2024-01-08
[16.0][FIX] product_variant_name: pass on variant name to template when we create the new product variant

JordiBForgeFlow:
The downside of this approach is that it will slow down the mass product creation as it calls to the product creation for each product individually.

### DIFF
--- a/product_variant_name/models/product_product.py
+++ b/product_variant_name/models/product_product.py
@@ -2,10 +2,22 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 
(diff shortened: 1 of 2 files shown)


JordiBForgeFlow:
@LoisRForgeFlow

JordiMForgeFlow (approved):
Similar approach as we had in product_variant_configurator module before https://github.com/OCA/product-variant/pull/325. LGTM

alexbottemanne (approved):

ThomasBinsfeld (approved):

Issue #330, opened 2024-01-10
[16.0] product_variant_configurator: field 'product_id' in module 'product.configurator' breaks inventory valuation report


JordiBForgeFlow:
After installing product_variant_configurator, if you go to 'Inventory > Reporting > Valuation' the following error message appears:

```
  File "/opt/odoo/odoo/addons/base/models/ir_ui_view.py", line 1318, in _postprocess_tag_groupby
    self._postprocess_view(node, field.comodel_name, editable=False, parent_name_manager=name_manager)
  File "/opt/odoo/odoo/addons/base/models/ir_ui_view.py", line 1144, in _postprocess_view
    if model_name not in self.env:
  File "/opt/odoo/odoo/api.py", line 538, in __contains__
    return model_name in self.registry
  File "/usr/lib/python3.10/_collections_abc.py", line 830, in __contains__
    self[key]
RecursionError: maximum recursion depth exceeded
```
Seems that the problem is caused by the field 'product_id' in the model 'product.configurator'. https://github.com/OCA/product-variant/blob/16.0/product_variant_configurator/models/product_configurator.py#L34

If we remove the field, the problem disappears.
If we remove this line from the stock valuation tree view, https://github.com/odoo/odoo/blob/16.0/addons/stock_account/views/stock_valuation_layer_views.xml#L57 the problem disappears.

So, it seems to me that when odoo tries to render the <groupby name="product_id"> it messes up because the model 'product.product' now has a field 'product_id' (inherited from 'product.configurator').

Perhaps the simplest approach is to rename the field 'product_id' to 'product_variant_id'.

cc @pedrobaeza

pedrobaeza:
Not so direct I'm afraid, as the configurator is inherited by for example `sale.order.line`, and the `product_id` field is put on purpose. What we can try to do is to parametrize the field to use for the "variant", defaulting to `product_id`, but possible to be changed per model inheriting the mixin, and see if we can change all the computes to use the proper dependency.

JordiBForgeFlow:
![image](https://github.com/OCA/product-variant/assets/7683926/27255bf1-8918-4842-930f-9f4f4d86efbb)

JordiBForgeFlow:
@pedrobaeza The problem is that the `product.product` model cannot have a field 'product_id' defined. So, it should not be inherited from `product.configurator` in the first place. Another option is to define two mixins.

pedrobaeza:
[...]

PR #331, opened 2024-01-11
[16.0][FIX] product_variant_configurator: Solve recursion error.

JordiBForgeFlow:
This fix solves a recursion problem caused when the groupby refers to a field of a many2one model, and that model also has the same field name. In this example, the stock.valuation.layer.tree view contains a groupby product_id, and the product.product model already has a product_id field, relating to the same model. Without this code we would reach a recursion error.

Fixes https://github.com/OCA/product-variant/issues/330

cc @JordiMForgeFlow @marinaaforgeflow @lauracforgeflow @pedrobaeza

### DIFF
--- a/product_variant_configurator/models/__init__.py
+++ b/product_variant_configurator/models/__init__.py
@@ -6,3 +6,4 @@ from . import product_configurator_attribute
 from . import product_product
 from . import product_template
 from . import product_template_attribute_line
+from . import ir_ui_view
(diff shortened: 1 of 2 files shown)


JordiMForgeFlow (approved):
Makes sense 👍🏼

MarinaAForgeFlow (approved):
LGTM

pedrobaeza (approved):
Great hack for avoiding to do other massive changes, and well documented for not having future doubts about its usage. The test coverage is limited though, so this can fail again in the future (or future versions), but it will require a tour test for recreating the condition, so let's continue as is.

/ocabot merge patch

JordiBForgeFlow:
@pedrobaeza thanks for your review.

PR #300, opened 2023-07-21
[FIX] product_variant_default_code: code_prefix in reference_mask

RabbitJon-S73:
Cherry-picked from v14 https://github.com/OCA/product-variant/pull/293

PR #338, opened 2024-03-12
[15.0][FIX] product_variant_default_code: recurrent prefix

chienandalu:
With a non automatic reference mask that has a code prefix, when that reference mask is recomputed, we could be repeating the prefix without noticing it. For instance, when we change an product.attribute name and that triggers all the default codes recomputations.

cc @Tecnativa TT48322

please review @pedrobaeza @victoralmau

### DIFF
--- a/product_variant_default_code/models/product.py
+++ b/product_variant_default_code/models/product.py
@@ -124,7 +124,13 @@ class ProductTemplate(models.Model):
             if automask or not rec.reference_mask:
                 rec.reference_mask = rec._get_default_mask()
             elif not automask and rec.code_prefix:
-                rec.reference_mask = rec.code_prefix + rec.reference_mask
+                reference_mask = rec.reference_mask
+                # Avoid prefixing the mask twice (or more).
+                # TODO: This needs a better design with a third field that sums both
+                # or using the sum of them in the variant default code computation
+                if reference_mask.startswith(rec.code_prefix):
+                    reference_mask = reference_mask.lstrip(rec.code_prefix)
+                rec.reference_mask = rec.code_prefix + reference_mask
 
     def _inverse_reference_mask(self):
         self._compute_reference_mask()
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Do you have deterministic steps to reproduce the problem? Are they reflected in the test as regression case?

chienandalu:
> Do you have deterministic steps to reproduce the problem? Are they reflected in the test as regression case?

Yes, it's stated in the description (non automatic mask template and a code prefix and having a trigger in the mask recomputation, like changing an attribute name) and a new test step is added to assert the fix

pedrobaeza:
Well, that are the conditions to trigger the problem, but not the steps in the sense of (create a product this way, do this and do that). Pre-commit is red.

Kev-Roche:
Is it the same problem describe here ? https://github.com/OCA/product-variant/pull/293 
If so, a cherry pick could be enough like here https://github.com/OCA/product-variant/pull/300 ?

chienandalu:
Thanks for the heads up @Kev-Roche 

I think in any case that those fixes could be problematic, as it could happen that the prefix could match with a part inside the reference_mask. That's why I used `startswith` so I could restrict the issue as much as possible.

Moreover, what if the mask should be legitly updated? That case would be ignored!

I'm going to add a test for such case

Addressing @pedrobaeza request for steps:

In a product template with variants (i.e.: Conference Chair):

- You'll have a reference mask ([Legs])
- Add a prefix: cc

**--> The prefix is duplicated in the reference mask: cccc[Legs]**

- Now you can delete de prefix manually from the mask and the resulting mask will be right: cc[Legs]
- Now change the name of the attribute Legs to something else.

**--> The mask computation is retriggered and we get a duplicated prefix again cccc[Legs].**

An extra issue would be that the mask isn't recomputed with the new attribute name, although that would need some extra work.

pedrobaeza:
Please reflect that steps to reproduce in the commit message, and it's good to go.

chienandalu:
Commit message amended

chienandalu:
CI Green

pedrobaeza (approved):

pedrobaeza:
@Kev-Roche is it OK for you?

Kev-Roche (approved):
LGTM

PR #346, opened 2024-04-12
[16.0][FIX] product_variant_configurator: change instance checking

Mantux11:
Creating a new Sales Order and selecting a product in the SO line, and choosing any subscription period results in a technical error.

Checking instance (`_name`) should not be on a specific record, it could be done on all products. And if there is no products, getting to first record raises and `Index out of range` error.

This fixes `Index out of range` error.


Du-ma (approved):

Du-ma:
@carolinafernandez-tecnativa 
Mind having a look since you migrated this to `16.0` and all.

pedrobaeza (approved):
/ocabot merge patch

PR #353, opened 2024-05-16
[16.0][MIG] sale_variant_configurator: Migration to version 16.0

carolinafernandez-tecnativa:
- Standard migration from 13 to 16.

@Tecnativa
TT46599

@pedrobaeza @pilarvargas-tecnativa


sale_variant_configurator/models/sale_order.py
```diff
@@ -168,7 +175,7 @@ def _onchange_product_attribute_ids_configurator(self):
     @api.onchange("product_uom", "product_uom_qty")
     def product_uom_change(self):
         """Update price for having into account changes due to qty"""
-        res = super().product_uom_change()
-        if not self.product_id:
+        res = {}
```
pilarvargas-tecnativa: Have you made sure that this method no longer exists to remove the call to the parent method?

carolinafernandez-tecnativa: Yes, no longer exists on v16

sale_variant_configurator/models/sale_order.py
```diff
@@ -63,20 +64,19 @@ def create(self, vals):
 
     @api.onchange("product_tmpl_id")
     def _onchange_product_tmpl_id_configurator(self):
-        res = super()._onchange_product_tmpl_id_configurator()
+        res = super()._onchange_product_tmpl_id_configurator() or {}
```
pilarvargas-tecnativa: I find it a bit weird to call super or {}

carolinafernandez-tecnativa: I added because sometimes, res returns None, so when it tries to add domain = res.setdefault("domain", {}), gives an error message as res is not inicialized

sale_variant_configurator/models/sale_order.py
pilarvargas-tecnativa: Then maybe you should rename it
https://github.com/OCA/sale-workflow/blob/16.0/sale_blanket_order/models/sale_orders.py#L115

Maybe this could have a conflict in some cases?

pedrobaeza: Now you have to override the computed methods instead.

carolinafernandez-tecnativa: Ready for review

pilarvargas-tecnativa:
Have you tried running the tests locally?

I did the test on a clean bd and I get the following error:

```
2024-05-22 11:27:27,567 1 ERROR devel odoo.addons.sale_variant_configurator.tests.test_sale_order: FAIL: TestSaleOrder.test_onchange_product_attribute_ids
  File "/opt/odoo/auto/addons/sale_variant_configurator/tests/test_sale_order.py", line 145, in test_onchange_product_attribute_ids
    self.assertEqual(line.price_unit + self.ptav_1.price_extra, 110)
AssertionError: 120.0 != 110
```

sale_variant_configurator/models/sale_order.py
```diff
@@ -102,28 +103,26 @@ def _onchange_product_tmpl_id_configurator(self):
...
-                product_tmpl.price,
+                product_tmpl.list_price,
                 product_tmpl.taxes_id,
                 self.tax_id,
             )
         return res
 
     @api.onchange("product_id")
-    def product_id_change(self):
+    def _onchange_product_id(self):
```
pilarvargas-tecnativa: You should not rename the method and create a new one, but hook into the existing one that has already been renamed.
In v15: https://github.com/odoo/odoo/blob/15.0/addons/sale/models/sale_order_line.py#L617
In v16: https://github.com/odoo/odoo/blob/16.0/addons/sale/models/sale_order_line.py#L939

Commit: https://github.com/odoo/odoo/commit/1d3b9f6be6f664eaa344028a96926f15fc59428f

carolinafernandez-tecnativa:
> + self.ptav_1.price_extra

Could you please try again? I have checked and with this new changes in local with new database this issue is fixed, thanks!

sale_variant_configurator/models/sale_order.py
```diff
@@ -102,28 +103,26 @@ def _onchange_product_tmpl_id_configurator(self):
...
-    def product_id_change(self):
+    def _onchange_product_id_warning(self):
         """Call again the configurator onchange after this main onchange
         for making sure the SO line description is correct.
 
         It also puts the proper lang in context for getting the product and
         attributes in the customer language.
         """
         obj = self.with_context(lang=self.order_id.partner_id.lang)
-        res = super(SaleOrderLine, obj).product_id_change()
```
pilarvargas-tecnativa: `res = super(SaleOrderLine, obj)._onchange_product_id_warning()`

sale_variant_configurator/models/sale_order.py
```diff
@@ -102,28 +103,26 @@ def _onchange_product_tmpl_id_configurator(self):
...
         obj._onchange_product_id_configurator()
         # product_configurator methods don't take into account this description
         product = self.product_id.with_context(lang=self.order_id.partner_id.lang)
         if product.description_sale:
             self.name = (self.name or "") + "\n" + product.description_sale
-        return res
```
pilarvargas-tecnativa: `return res`

pedrobaeza:
On v16, AFAIK, there are no more onchanges in `sale.order.line`. All are computed writable fields, so you shouldn't insist in making this to work with onchanges, but to adapt the mechanism overriding computed methods.

carolinafernandez-tecnativa:
> On v16, AFAIK, there are no more onchanges in `sale.order.line`. All are computed writable fields, so you shouldn't insist in making this to work with onchanges, but to adapt the mechanism overriding computed methods.



> On v16, AFAIK, there are no more onchanges in `sale.order.line`. All are computed writable fields, so you shouldn't insist in making this to work with onchanges, but to adapt the mechanism overriding computed methods.

The only one missing was _compute_name, could you please check it now?

product_variant_sale_price/models/product_product.py
```diff
@@ -44,7 +45,12 @@ def _get_combination_info(
             parent_combination,
             only_template,
         )
-        res["price_extra"] = 0.0
+        test_condition = not config["test_enable"] or (
```
rousseldenis, line 48: Try to avoid such things. That's programmatically incorrect.

victoralmau: IMO, in some cases, this behavior is absolutely wrong and you need that compatibility to avoid breaking tests.

When running the pipelines in gitlab, `product_variant_sale_price` is installed and the "_expected_" price is different from what the `sale_variant_configurator` tests expect to have (not all of us need the `product_variant_sale_price` module).

This technique is often used when there is no other "_possible solution_" to resolve incorrect test data.

product_variant_configurator/models/product_configurator.py
```diff
@@ -48,6 +49,7 @@ class ProductConfigurator(models.AbstractModel):
     )
     def _compute_can_be_created(self):
         for rec in self:
+            rec.price_extra = sum(rec.mapped("product_attribute_ids.price_extra"))
```
pedrobaeza: A computed can't assign other value outside of their definition.

product_variant_configurator/models/product_configurator.py
```diff
@@ -147,15 +147,14 @@ def _onchange_product_attribute_ids_configurator(self):
         if not self.product_id:
             product_tmpl = self.product_tmpl_id
             values = self.product_attribute_ids.mapped("value_id")
-            if "partner_id" in self._fields:
+            if self._partner_id_field in self._fields:
```
pedrobaeza, line 154: I don't think we should use these onchange helpers in v16, as `sale` is now using computed writable for everything. Better to overwrite that computed fields instead.

sale_variant_configurator/models/sale_order.py
```diff
@@ -48,127 +50,30 @@ class SaleOrderLine(models.Model):
...
-
-    @api.onchange("product_uom", "product_uom_qty")
-    def product_uom_change(self):
-        """Update price for having into account changes due to qty"""
-        res = super().product_uom_change()
-        if not self.product_id:
-            self._update_price_configurator()
-        return res
+    @api.depends("product_attribute_ids")
```
pedrobaeza, line 77: Document that is for adding the dependency

product_variant_configurator/models/product_configurator.py
victoralmau: Once again I don't understand what you want to do here.
This code is in `product_variant_configurator`, and it is used to re-define the `name` field (according to the partner language).
This method is used for `sale_variant_configurator` and `purchase_variant_configurator` to do this correctly.

pedrobaeza: OK, don't worry. Let's continue this way.

pedrobaeza (approved):
/ocabot merge nobump

PR #354, opened 2024-05-16
[16.0][MIG] sale_product_variant_attribute_tax: Migration to version 16.0

carolinafernandez-tecnativa:
- Depends on: https://github.com/OCA/product-variant/pull/353
- sale_variant_configurator dependency needed otherwise getting error message product_attribute_lines key error.
- Remove sale dependency as sale_variant_configurator already has sale dependency

@Tecnativa
TT46598

@pedrobaeza @pilarvargas-tecnativa


sale_product_variant_attribute_tax/models/sale_order.py
```diff
@@ -10,7 +10,7 @@ class SaleOrderLine(models.Model):
     def _onchange_product_attribute_ids_configurator(self):
         res = super()._onchange_product_attribute_ids_configurator()
         # if there's a product set, taxes on that product have priority
-        if self.product_id or not self.product_tmpl_id:
+        if not self.product_tmpl_id:
```
victoralmau: Why this change is necessary?

sale_product_variant_attribute_tax/tests/test_sale_product_variant_attribute_tax.py
```diff
@@ -57,10 +58,13 @@ def setUpClass(cls):
...
             }
         )
+        cls.product = cls.env["product.product"].create(
+            {"name": "Test template", "product_tmpl_id": cls.product_template.id}
+        )
         order_form = Form(cls.env["sale.order"])
         order_form.partner_id = cls.partner
         with order_form.order_line.new() as order_line_form:
-            order_line_form.product_tmpl_id = cls.product_template
+            order_line_form.product_id = cls.product_template.product_variant_id
```
victoralmau: Why this change is necessary?

victoralmau (changes requested):

pedrobaeza:
cc @carolinafernandez-tecnativa

sale_product_variant_attribute_tax/models/sale_order.py
carolinafernandez-tecnativa: If product is set we need to recalculate

sale_product_variant_attribute_tax/tests/test_sale_product_variant_attribute_tax.py
carolinafernandez-tecnativa: There is a sale_order_line constraint  sale_order_line_accountable_required_fields thats need product_jd completed, when product_tmpl_id is set, product_id is null.

pedrobaeza:
@victoralmau is it OK for you?

sale_product_variant_attribute_tax/models/sale_order.py
victoralmau: IMO the above behavior is correct or we will always define the taxes of the attribute omitting completely those of the product template, the above behavior causes it to be defined only when there is no `product_id` defined yet (the template has several attributes and only some of them have been defined but not all of them yet).

sale_product_variant_attribute_tax/tests/test_sale_product_variant_attribute_tax.py
```diff
@@ -57,10 +58,29 @@ def setUpClass(cls):
...
+                        {
+                            "product_tmpl_id": cls.product_template.id,
+                            "attribute_id": cls.attribute.id,
+                            "value_id": cls.attribute_value.id,
+                            "owner_model": "sale.order.line",
+                        },
+                    )
+                ],
+            }
+        )
```
victoralmau: This is not necessary, you can remove `"no_create_variants": "yes"` first.

It is important to remember the behavior of `sale_variant_configurator`, do NOT manually define `product_id`, we define a template (`product_tmpl_id`) and attribute values and the `product_id` field will define itself.

sale_product_variant_attribute_tax/models/sale_order.py
pedrobaeza: As it was, it was working correctly, so better to not change it unless something has changed across the versions.

victoralmau:
Thinking again about this functionality, I don't understand not included in `sale_variant_configurator` (is probably related to i don't understand the need for `product_variant_attribute_tax` because you could always use the template taxes, but that's another debate).

pedrobaeza:
Let me explain this feature, that is somehow independent of `sale_variant_configurator`, but right now it's linked for not doing 2 modules.

The question is that the product taxes are defined at template level, so you can't have some variants with an specific tax. The use case of our customer is that one attribute makes that all the variants with such attribute has different taxes. This module + the other one covers this gap adding the taxes at attribute level for having to define it only once, instead of searching for other solution like having taxes at template level that forces you to define the specific taxes for each of the variants that are affected.

sale_product_variant_attribute_tax/__manifest__.py
```diff
@@ -1,14 +1,15 @@
...
     "summary": "Bring the taxes associated to product values",
-    "version": "13.0.1.0.0",
+    "version": "16.0.1.0.0",
     "development_status": "Production/Stable",
     "category": "Sales/Sales",
     "website": "https://github.com/OCA/product-variant",
     "author": "Tecnativa, Odoo Community Association (OCA)",
     "license": "AGPL-3",
-    "depends": ["product_variant_attribute_tax", "sale_variant_configurator"],
+    "depends": ["sale_variant_configurator", "product_variant_attribute_tax"],
```
pedrobaeza: Don't modify this order. It was even in alphabetical order, which is the desirable thing.

sale_product_variant_attribute_tax/models/sale_order.py
pedrobaeza: I confirm you shouldn't modify this. Please leave it as it was.

sale_product_variant_attribute_tax/tests/test_sale_product_variant_attribute_tax.py
pedrobaeza: Exactly.

pedrobaeza (changes requested):
- Please squash the first 3 commits into one.
- You must split your migration into 2 commits: the pre-commit one + the real migration one.
- Attend the rest of the inline comments.

victoralmau:
> Let me explain this feature, that is somehow independent of `sale_variant_configurator`, but right now it's linked for not doing 2 modules.
> 
> The question is that the product taxes are defined at template level, so you can't have some variants with an specific tax. The use case of our customer is that one attribute makes that all the variants with such attribute has different taxes. This module + the other one covers this gap adding the taxes at attribute level for having to define it only once, instead of searching for other solution like having taxes at template level that forces you to define the specific taxes for each of the variants that are affected.

Ok, now I understand the requirement, thanks for clarifying.

pedrobaeza:
@carolinafernandez-tecnativa can you please finish this. You can include my previous reasoning as CONTEXT.rst in the readme.

sale_product_variant_attribute_tax/models/sale_order.py
carolinafernandez-tecnativa: Please check without this condition if not self.product_tmpl_id, the module is not working properly.

carolinafernandez-tecnativa:
> @carolinafernandez-tecnativa can you please finish this. You can include my previous reasoning as CONTEXT.rst in the readme.

done

carolinafernandez-tecnativa:
> * Please squash the first 3 commits into one.
> * You must split your migration into 2 commits: the pre-commit one + the real migration one.
> * Attend the rest of the inline comments.

pending to do

carolinafernandez-tecnativa:
@victoralmau could you please help me with this?thanks!

victoralmau (approved):

carolinafernandez-tecnativa:
ping @pedrobaeza

sale_product_variant_attribute_tax/readme/CONTEXT.rst
```diff
@@ -0,0 +1,9 @@
+This module is somehow independent of sale_variant_configurator, but right now it's linked for not doing 2 modules.
```
pedrobaeza: This text is not suitable to be put here as it is. You have to work around it a bit. Expressions like "The question is that" or "This module + the other one" are for you, not for being reflected in a README.

Also, the explanation "This module is somehow independent of sale_variant_configurator, but right now it's linked for not doing 2 modules." is more for ROADMAP, not for here (and also to be adapted).

carolinafernandez-tecnativa: Could you please give me a suggestion for adding as an explanation?

pedrobaeza: For CONTEXT.rst:

```
As the product taxes are defined at product template level, you can't have some variants with an specific tax.

These modules family covers the gap, adding the taxes at attribute level. This design decision has been taken for allowing to define the tax only once, as other options like having taxes at product variant level will force you to define the specific taxes for each of the variants that are affected.
```

For ROADMAP.rst:

```
- This module requires `sale_variant_configurator` as a way to avoid 2 glue modules, but it may be independent.
```
```

carolinafernandez-tecnativa:
ping @pedrobaeza

pedrobaeza (approved):
/ocabot merge nobump

PR #363, opened 2024-06-27
[17.0][FIX] product_variant_sale_price: edit Sales Price

gabe-grinspan:
There is an issue in 17.0 where the Sales Price field is not editable in the quick view page found by navigating to variants from a parent product page. This module makes it such that the field is separated between products and variants. It should be easy to modify this field without having to navigate to the variants list page and search.

I believe this worked in previous versions, as the code seems to indicate there was a purpose for clearing `attrs` but in Odoo 17, Odoo removed that attribute and replaced it with options. As it is, options has nothing to do with `readonly` in this case.

### DIFF
--- a/product_variant_sale_price/README.rst
+++ b/product_variant_sale_price/README.rst
@@ -70,6 +70,7 @@ Contributors
 -  RabbitJon-S73 <roger@studio73.es>
 -  Emanuel Cino <ecino@compassion.ch>
 -  Pedroguirao <pguirao@puntsistemes.es>
+-  Gabriel Grinspan <gabriel@rooteam.net>
 
 Maintainers
 -----------
--- a/product_variant_sale_price/readme/CONTRIBUTORS.md
+++ b/product_variant_sale_price/readme/CONTRIBUTORS.md
@@ -6,3 +6,4 @@
 - RabbitJon-S73 \<<roger@studio73.es>\>
 - Emanuel Cino \<<ecino@compassion.ch>\>
 - Pedroguirao \<<pguirao@puntsistemes.es>\>
+- Gabriel Grinspan \<<gabriel@rooteam.net>\>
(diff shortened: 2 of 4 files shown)


gabriel-grinspan:
@HaraldPanten
Are you the right person to merge this? Please review.

gabriel-grinspan:
@pedrobaeza

product_variant_sale_price/__manifest__.py
```diff
@@ -8,7 +8,7 @@
     "version": "17.0.1.0.0",
     "category": "Product Management",
     "website": "https://github.com/OCA/product-variant",
-    "author": "Tecnativa, " "Odoo Community Association (OCA)",
+    "author": "Tecnativa, Rooteam, " "Odoo Community Association (OCA)",
```
pedrobaeza: This change doesn't qualify for co-authorship, so remove it.

pedrobaeza (changes requested):
Follow commit message guidelines: https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message

product_variant_sale_price/__manifest__.py
gabriel-grinspan: Do I get automatically added as a contributor, or do I have to manually update the README?

gabriel-grinspan:
@pedrobaeza 
This should be good now.

product_variant_sale_price/__manifest__.py
pedrobaeza: The commit attribution is already in GitHub, but for being explicitly in the README file, you have to add yourself in https://github.com/OCA/product-variant/blob/17.0/product_variant_sale_price/readme/CONTRIBUTORS.md and execute `pre-commit`

pedrobaeza: Do I merge as is for this time?

gabriel-grinspan:
@pedrobaeza 
I have amended my commit, updated the contributors, and run the pre-commit.

pedrobaeza (approved):
OK, let's fast-track as very clear:

/ocabot merge minor

PR #366, opened 2024-09-06
[16.0][ADD] product_variant_change_template

ChrisOForgeFlow:
This tool allow reassign existing variants to an empty template, this ideal when odoo has template creation as a initial approach, but after some time you decide to combine similar products as a variants belong a template

You should create empty variants with attributes, then you can select variant and change template

When template original empty variants, automatic inactivated by wizard

Restrictions:

- You can't assign if original uom is different of destination uom


AlexPForgeFlow:
Can we reopen?

PR #372, opened 2024-11-02
[16.0][FIX] product_variant_default_code ir.config_parameter

giarve:
image

this setting is useless because module name is not considered

for now is has to be set manually on ir.config_parameter without prefix


rousseldenis (approved):
Indeed. This is weird nobody saw that before

marcelsavegnago (approved):

giarve:
too bad this was closed, it is an important fix

pedrobaeza (approved):
/ocabot merge patch

PR #378, opened 2024-12-11
[16.0][IMP] sale_variant_configurator: Improve Error 

unaiberis:
ValueError: Wrong @depends on '_compute_price_unit' (compute method of field sale.order.line.price_unit). Dependency field 'product_attribute_ids' not found in model sale.order.line.

### DIFF
--- a/sale_variant_configurator/models/sale_order.py
+++ b/sale_variant_configurator/models/sale_order.py
@@ -74,7 +74,6 @@ class SaleOrderLine(models.Model):
             res = (res or "") + "\n" + product.description_sale
         return res
 
-    @api.depends("product_attribute_ids")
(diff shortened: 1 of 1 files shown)


sale_variant_configurator/models/sale_order.py
```diff
@@ -74,7 +74,6 @@ def _get_product_description(self, template, product, product_attributes):
             res = (res or "") + "\n" + product.description_sale
         return res
 
-    @api.depends("product_attribute_ids")
```
rousseldenis, line 77: @unaiberis Are you sure about this?

https://github.com/OCA/product-variant/blob/16.0/product_variant_configurator/models/product_configurator.py#L22

PR #394, opened 2025-02-05
[18.0][MIG] product_attribute_value_archive: Migration to 18.0

LucasTran380381:
* Product attribute value has active field 
image
* when delete attribute module will archive instead of unlink
image


product_attribute_value_archive/models/product_attribute_value.py
```diff
@@ -18,7 +18,7 @@ def _get_related_variants(self):
         product.
         """
         self.ensure_one()
-        self.flush()
+        self.flush_recordset()
```
ivantodorovich: Actually, here, we must flush all fields used in the query... and not only those of the recordset.

I think we need to do something like this:

```py
self.env["product.product"].flush_model(
    [
        "product_template_attribute_value_ids",
        "product_template_variant_value_ids",
    ]
)
self.env["product.template.attribute.value"].flush_model(["product_attribute_value_id"])
```

product_attribute_value_archive/models/product_attribute_value.py
```diff
@@ -47,6 +47,7 @@ def _get_pav_to_archive(self):
             related_variants = pav._get_related_variants()
             # Archive only if all related variants are archived
             # (none is active)
+            related_variants.fetch(["active"])
```
ivantodorovich: Why is this explicit fetch needed here? 🤔

ivantodorovich (changes requested):
Thanks!

product_attribute_value_archive/models/product_attribute_value.py
LucasTran380381: mapped will get all field in table "product.product"
image

we will fetch only field `active`
image

ivantodorovich: So this is a performance optimization to no get prefetch all fields.
Similar to `related_variants.with_context(prefetch_fields=False).mapped('active')`

Perhaps keep an inline comment explaining the purpose

ivantodorovich (approved):

ajaniszewska-dev (approved):

rousseldenis (approved):

hparfr:
Why `auto_install = True `  ? (I know it was like that in v13).

PR #395, opened 2025-02-24
[16.0][FIX] sale_variant_configurator: Unit of measure in product templates

victoralmau:
Unit of measure in product templates

Steps to reproduce the error:
- Go to Sales > Configuration and activate 'Units of Measure'.
- Go to a quotation and add a line indicating a Product Template with several attributes (Conference Chair for example).
- You must be able to select a unit of measure + the unit of measure is set automatically.

Other versions (v13 for example) use another technique https://github.com/OCA/product-variant/blob/2922137d2de670010200fa0cb7cdf1469bffafbe/sale_variant_configurator/models/sale_order.py#L69

**Before**:
antes

**After**:
despues

Please @pedrobaeza and @pilarvargas-tecnativa can you review it?

@Tecnativa TT55223

### DIFF
--- a/sale_variant_configurator/models/sale_order.py
+++ b/sale_variant_configurator/models/sale_order.py
@@ -31,6 +31,13 @@ class SaleOrderLine(models.Model):
         string="Product Template (no related)",
     )
     product_id = fields.Many2one(required=False)
+    product_uom_category_id = fields.Many2one(
+        comodel_name="uom.category",
+        compute="_compute_product_uom_category_id",
+        # We need to define related=False so that the field is only compute
+        # and not related.
+        related=False,
+    )
 
     _sql_constraints = [
         (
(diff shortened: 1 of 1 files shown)


sale_variant_configurator/models/sale_order.py
```diff
@@ -31,6 +31,11 @@ class SaleOrderLine(models.Model):
         string="Product Template (no related)",
     )
     product_id = fields.Many2one(required=False)
+    product_uom_category_id = fields.Many2one(
+        comodel_name="uom.category",
+        compute="_compute_product_uom_category_id",
+        related=False,
```
pedrobaeza, line 39: Why are you removing this as related? If it's because the related is not "acting on time", please try to look for another option, as this can have side effects.

victoralmau: We define `related=False` to avoid it being a related (and therefore a compute). Another possible alternative option is to create another `product_tmpl_uom_category_id` field and modify the domain of the `product_uom` field, what do you prefer?

pedrobaeza: OK, now I see the problem seeing https://github.com/OCA/OCB/blob/6c40bebb4eb0ca9566378ac7b82b50f9c1fbb0bc/addons/sale/models/sale_order_line.py#L94 when no `product_id` is still defined.

sale_variant_configurator/models/sale_order.py
```diff
@@ -50,6 +55,27 @@ class SaleOrderLine(models.Model):
...
+        return super(SaleOrderLine, self - lines_with_template)._compute_product_uom()
+
+    @api.depends("product_tmpl_id", "product_id")
+    def _compute_product_uom_category_id(self):
+        for line in self:
+            product = line.product_id or line.product_tmpl_id
+            if product:
+                line.product_uom_category_id = product.uom_id.category_id
+            else:
+                line.product_uom_category_id = line.product_uom_category_id
```
pedrobaeza, line 85: Is this still needed in 16.0?

sale_variant_configurator/models/sale_order.py
```diff
@@ -50,6 +55,27 @@ class SaleOrderLine(models.Model):
...
 
+    @api.depends("product_tmpl_id")
+    def _compute_product_uom(self):
+        lines_with_template = self.filtered(
+            lambda x: x.product_tmpl_id and not x.product_id
+        )
+        for line in lines_with_template:
+            if not line.product_uom or (
+                line.product_tmpl_id.uom_id.id != line.product_uom.id
+            ):
```
pedrobaeza, line 70: Is this comparison due to any reason? If so, please put a comment saying the reason for not being removed in the future.

sale_variant_configurator/models/sale_order.py
victoralmau: Yes, it is a compute no store and it is necessary that all records have a value to avoid the CacheMis error.

sale_variant_configurator/models/sale_order.py
victoralmau: Added comment to clarify.

sale_variant_configurator/models/sale_order.py
victoralmau: Added comment to clarify.

pedrobaeza (approved):
We can fast-track this for fixing the issue when no product is determined. I think the impact in this one hasn't been found on standard, as most of the times the variants are pre-generated.

/ocabot merge patch

PR #396, opened 2025-02-24
[16.0][FIX] purchase_variant_configurator: Unit of measure in product templates

victoralmau:
Unit of measure in product templates

Steps to reproduce the error:
- Adds the 'Manage Multiple Units of Measure' permission to the user.
- Go to a quotation and add a line indicating a Product Template with several attributes (Conference Chair for example).
- You must be able to select a unit of measure.

**Before**:
antes

**After**:
despues

Please @pedrobaeza and  @pilarvargas-tecnativa can you review it?

@Tecnativa TT55224


pedrobaeza:
Apply the same comments as in sale.

pedrobaeza (approved):
/ocabot merge patch

PR #398, opened 2025-02-26
[16.0][FIX] sale_variant_configurator: Do not create products for downpayments

victoralmau:
Do not create products for downpayments

**Before**:
antes

**After**:
despues

@Tecnativa


pedrobaeza (approved):
Obvious patch:

/ocabot merge patch

PR #399, opened 2025-03-03
[16.0][FIX] puchase_variant_configurator: Avoid error when similar sellers found

pedrobaeza:
The standard code for selecting the seller (product.supplierinfo) performs a sorted over sequence, min_qty, price and ending with id. When calling it from this module with a virtual record, as we do not having yet the variant created, .id gets NewId type records (even if the supplierinfo exists), so we get the error:

```
TypeError: '<' not supported between instances of 'NewId' and 'NewId'
```

For avoiding this, let's remove the id from the sorting criteria as a workaround.

@Tecnativa TT55224

### DIFF
--- a/purchase_variant_configurator/models/product_supplierinfo.py
+++ b/purchase_variant_configurator/models/product_supplierinfo.py
@@ -34,3 +34,14 @@ class ProductSupplierinfo(models.Model):
         return super().search(
             args, offset=offset, limit=limit, order=order, count=count
         )
+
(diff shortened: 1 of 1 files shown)


victoralmau:
Reviewing in runboat the use case the error still occurs.

pedrobaeza:
The context is not propagated to that method, so we need to open a bit the scope of the patch. Done.

victoralmau (approved):
Functional review OK

PR #400, opened 2025-03-20
[18.0][FIX] product_attribute_value_archive: Fix issue where attribute_id may not be set

natuan9:
ISSUE: In case product.attribute.value is created directly from the product form view, values parameter in create method doesn't contain attribute_id.
This PR resolves this issue

image

### DIFF
--- a/product_attribute_value_archive/models/product_attribute_value.py
+++ b/product_attribute_value_archive/models/product_attribute_value.py
@@ -69,12 +69,16 @@ class ProductAttributeValue(models.Model):
         values_to_create = []
         unarchived_record_ids = set()
         for value in values:
+            attribute_id = value.get(
+                "attribute_id", self._context.get("default_attribute_id")
+            )
             existing_archived_value = self.search(
                 [
                     ("active", "=", False),
                     ("name", "=", value["name"]),
-                    ("attribute_id", "=", value["attribute_id"]),
-                ]
+                    ("attribute_id", "=", attribute_id),
+                ],
+                limit=1,
             )
             if existing_archived_value:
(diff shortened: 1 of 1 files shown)


david-s73:
@mmequignon @rousseldenis Can you check and review the pr?

david-s73 (approved):

product_attribute_value_archive/models/product_attribute_value.py
```diff
@@ -69,12 +69,16 @@ def create(self, values):
         values_to_create = []
         unarchived_record_ids = set()
         for value in values:
+            attribute_id = value.get(
+                "attribute_id", self._context.get("default_attribute_id")
```
rousseldenis, line 73:
```suggestion
                "attribute_id", self.env.context.get("default_attribute_id")
```

`_context` is deprecated for years.

product_attribute_value_archive/models/product_attribute_value.py
```diff
@@ -69,12 +69,16 @@ def create(self, values):
...
+            )
             existing_archived_value = self.search(
                 [
                     ("active", "=", False),
                     ("name", "=", value["name"]),
-                    ("attribute_id", "=", value["attribute_id"]),
-                ]
+                    ("attribute_id", "=", attribute_id),
+                ],
+                limit=1,
```
rousseldenis, line 81: Why limiting to `1` in this case ?

david-s73:
@rousseldenis I have created a new pr with the changes you have commented in this one since it does not seem that I will review the changes you indicated.
https://github.com/OCA/product-variant/pull/405

PR #401, opened 2025-03-25
[16.0][ADD] product_variant_configurator_change_template

AlexPForgeFlow:
Superseeds: https://github.com/OCA/product-variant/pull/366

This tool allow reassign existing variants to an empty template, this ideal when odoo has template creation as a initial approach, but after some time you decide to combine similar products as a variants belong a template

You should create empty variants with attributes, then you can select variant and change template

When template original empty variants, automatic inactivated by wizard

Restrictions:

- You can't assign if original uom is different of destination uom


JordiMForgeFlow:
@AlexPForgeFlow maybe it would make more sense for the module to be named like 'product_variant_configurator_change_template' as it relies on the 'product_variant_configurator' module. WDYT?

Could you also add the copyright and license lines in the module files?

Not sure if you already thought about it but it would probably make sense to also have a glue module with MRP to handle the changes on BoMs

product_variant_configurator_change_template/wizard/wizard_product_variant_change_template.py
```diff
@@ -0,0 +1,212 @@
+# Copyright 2025 ForgeFlow
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+import logging
+
+from odoo import Command, _, api, fields, models
+from odoo.exceptions import UserError
+
+_logger = logging.getLogger(__name__)
```
JordiMForgeFlow, line 9: unused, we could remove

zeroheure:
This is interesting. I often do that by hand on my wife's Odoo.

PR #385, opened 2025-01-17
[18.0][MIG] product_variant_configurator: Migration to 18.0

BhaveshHeliconia:

PR #404, opened 2025-05-19
[18.0][MIG] product_variant_configurator: Migration to 18.0

notsebtho:

PR #406, opened 2025-06-02
[18.0] [MIG] product_variant_configurator_manual_creation

marielejeune:
Migration from version 15.0

Addon used since v14 without issue -> change development status to Beta.

Adding unit tests to make codecov succeed

Depends on https://github.com/OCA/product-variant/pull/385


product_variant_configurator_manual_creation/wizards/wizard_product_variant_configurator_manual_creation_view.xml
```diff
@@ -0,0 +1,62 @@
...
+                        <field name="variants_to_create" />
+                    </group>
+                    <field name="line_ids" nolabel="1">
+                        <list editable="top" create="false" delete="false">
+                            <field name="attribute_id" readonly="1" force_save="1" />
+                            <field name="required" readonly="1" force_save="1" />
+                            <field
+                                name="attribute_value_ids"
+                                widget="many2many_tags"
+                                invisible="1"
```
CILC98:
```suggestion
                                column_invisible="1"
```

CILC98 (changes requested):

product_variant_configurator_manual_creation/wizards/wizard_product_variant_configurator_manual_creation_view.xml
marielejeune: Right, thanks. Done

marielejeune:
As for now this PR depends on https://github.com/OCA/product-variant/pull/404. It could depend on https://github.com/OCA/product-variant/pull/409 but this PR looks abandoned too...

marielejeune:
Changed dependency to https://github.com/OCA/product-variant/pull/385 as it is the oldest and more mature PR.

marielejeune:
@CILC98 I've fixed your suggestion and added a commit for improving unit tests and make codecov succeed. Can you re-review please?

CILC98:
@marielejeune  Let's hope change proposal #385  is finalized. Does that sound okay to you?

marielejeune:
> @marielejeune Let's hope change proposal #385 is finalized. Does that sound okay to you?

If Pedro's comments are addressed it should be ok and almost ready to merge I think

rousseldenis:
/ocabot migration product_variant_configurator_manual_creation

rousseldenis (approved):

marielejeune:
@CILC98 The `product_variant_configurator` module was merged. I've rebased this PR. Could you please update your review? We're almost there!

product_variant_configurator_manual_creation/wizards/wizard_product_variant_configurator_manual_creation.py
```diff
@@ -0,0 +1,163 @@
...
+                        "active": product_tmpl.active,
+                    }
+                )
+        if current_variants_to_activate:
+            variants_to_show |= current_variants_to_activate
+            current_variants_to_activate.write({"active": True})
+        if current_variants_to_create:
+            variants_to_show |= product_model.create(current_variants_to_create)
+        if variants_to_show:
+            action = self.env.ref("product.product_variant_action").read()[0]
```
CILC98:
```suggestion
            action = self.env.ref("product.product_variant_action").sudo().read()[0]
```

CILC98 (changes requested):

product_variant_configurator_manual_creation/wizards/wizard_product_variant_configurator_manual_creation.py
CILC98: @marielejeune Only the system administrator can read this model.

marielejeune: Right, changed

product_variant_configurator_manual_creation/wizards/wizard_product_variant_configurator_manual_creation.py
```diff
@@ -0,0 +1,163 @@
...
+            for line in self.line_ids
+            if line.selected_value_ids
+        ]
+        return list(selected_items and product(*selected_items))
+
+    @api.model
+    def default_get(self, fields_list):
+        values = super().default_get(fields_list)
+        values["product_tmpl_id"] = self.env.context.get("active_id")
+        return values
```
CILC98, line 53: Image

The button is in variants (which is useful), but in default_get you're trying to open it only from the product_template. You should check active_model.
```suggestion
    @api.model
    def default_get(self, fields_list):
        values = super().default_get(fields_list)
        product_tmpl_id = False
        if self._context.get("active_model") == "product.template":
            product_tmpl_id = self.env.context.get("active_id")
        elif self._context.get("active_model") == "product.product":
            product_tmpl_id = self.env["product.product"].browse(
                self.env.context.get("active_id")
            ).product_tmpl_id.id
        values["product_tmpl_id"] = product_tmpl_id
        return values
```

product_variant_configurator_manual_creation/wizards/wizard_product_variant_configurator_manual_creation.py
```diff
@@ -0,0 +1,163 @@
...
+
+    @api.onchange("product_tmpl_id")
+    def _onchange_product_tmpl(self):
```
CILC98, line 56: 1 month ago, I tried to open the wizard with a product with 2 attributes (each with 2000 values) and this method caused the wizard to never open.

I think the wizard should show all values ​​and only create the variants that were not created before.

What do you think about this?

CILC98 (changes requested):

CILC98:
@marielejeune

product_variant_configurator_manual_creation/wizards/wizard_product_variant_configurator_manual_creation.py
marielejeune: Indeed the method can be really slow for attributes with a lot of values.
Have you checked if the problem came from the computation of `self.product_tmpl_id._get_values_without_variant()` (for computing the variants still to create), or is it just the fact of creating the lines with all values in `attribute_value_ids`?
I'm afraid that even if we show all values, creating the lines will all 2000 values in each line, in field `attribute_value_ids`, will still be too heavy?

What do you think?

product_variant_configurator_manual_creation/wizards/wizard_product_variant_configurator_manual_creation.py
marielejeune: You're right. This bug was already present in v15, hence is independent of the migration. For more clarity I've fixed it in a new commit.
Edit: I must check why tests fail, I'll do it asap.

marielejeune: Hi, I've fixed the tests. The active_model must now be specified in tests. Moreover I've added a test to simulate the opening of the wizard from the product.product.

product_variant_configurator_manual_creation/wizards/wizard_product_variant_configurator_manual_creation.py
CILC98: The problem is with `self.product_tmpl_id._get_values_without_variant()`. I omitted this method and displayed all the values, and the wizard opened faster. But I think this is an exceptional case.

marielejeune: Thanks for your explanation.
I think the case of an attribute with 2000 values is quite exceptional, isn't it?
I would suggest to keep the same behavior as in v15 for the migration. Then you can open a new PR/issue if you want to discuss this improvement. What do you think?

CILC98: ok

CILC98 (approved):

StefanRijnhart (approved):
Thanks for the work @marielejeune, and thanks for the extensive review @CILC98!

/ocabot merge nobump

StefanRijnhart:
We're on collision here with `product_variant_name`, which introduces a *required* field product.product name and is going to break the tests of other modules in this repo as well sooner or later: https://github.com/OCA/product-variant/blob/18.0/product_variant_name/models/product_product.py. Can you demote it to a rebel module using include/exclude directives as is done here: https://github.com/OCA/web/commit/0b5aa6775dfcc46a183e142d3dfa7755b43fbd37.

<details>
<summary>Test failiure traceback</summary>

```
 2025-12-24 19:19:30,321 309 ERROR odoo odoo.addons.product_variant_configurator_manual_creation.tests.test_product_variant_configurator_manual_creation: ERROR: TestProductVariantConfiguratorManualCreation.test_product_attribute_manual_creation
  File "/__w/product-variant/product-variant/product_variant_configurator_manual_creation/tests/test_product_variant_configurator_manual_creation.py", line 95, in test_product_attribute_manual_creation
    variant_creation_wizard2.action_create_variants()
  File "/__w/product-variant/product-variant/product_variant_configurator_manual_creation/wizards/wizard_product_variant_configurator_manual_creation.py", line 128, in action_create_variants
    variants_to_show |= product_model.create(current_variants_to_create)
  File "<decorator-gen-237>", line 2, in create
  File "/opt/odoo/odoo/api.py", line 498, in _model_create_multi
    return create(self, arg)
  File "/__w/product-variant/product-variant/product_variant_configurator/models/product_product.py", line 175, in create
    return super().create(vals_list)
  File "<decorator-gen-170>", line 2, in create
  File "/opt/odoo/odoo/api.py", line 498, in _model_create_multi
    return create(self, arg)
  File "/opt/odoo/addons/product/models/product_product.py", line 376, in create
    products = super(ProductProduct, self.with_context(create_product_product=False)).create(vals_list)
  File "<decorator-gen-139>", line 2, in create
  File "/opt/odoo/odoo/api.py", line 498, in _model_create_multi
    return create(self, arg)
  File "/opt/odoo/addons/mail/models/mail_thread.py", line 293, in create
    threads = super(MailThread, self).create(vals_list)
  File "<decorator-gen-236>", line 2, in create
  File "/opt/odoo/odoo/api.py", line 498, in _model_create_multi
    return create(self, arg)
  File "/__w/product-variant/product-variant/product_variant_configurator/models/product_configurator.py", line 258, in create
    return super().create(vals_list)
  File "<decorator-gen-31>", line 2, in create
  File "/opt/odoo/odoo/api.py", line 498, in _model_create_multi
    return create(self, arg)
  File "/opt/odoo/odoo/models.py", line 5034, in create
    records = self._create(data_list)
  File "/opt/odoo/odoo/models.py", line 5218, in _create
    cr.execute(SQL(
  File "/opt/odoo/odoo/sql_db.py", line 357, in execute
    res = self._obj.execute(query, params)
psycopg2.errors.NotNullViolation: null value in column "name" violates not-null constraint
DETAIL:  Failing row contains (118, 69, 1, 1, null, null, null, null, null, null, t, null, 2025-12-24 19:19:30.215696, 2025-12-24 19:19:30.215696, null, null, f, null).

 
2025-12-24 19:19:30,400 309 ERROR odoo odoo.sql_db: bad query: b'INSERT INTO "product_product" ("active", "create_date", "create_uid", "manual_code", "product_tmpl_id", "write_date", "write_uid") VALUES (true, \'2025-12-24 19:19:30.215696\', 1, false, 70, \'2025-12-24 19:19:30.215696\', 1) RETURNING "id"'
ERROR: null value in column "name" violates not-null constraint
DETAIL:  Failing row contains (120, 70, 1, 1, null, null, null, null, null, null, t, null, 2025-12-24 19:19:30.215696, 2025-12-24 19:19:30.215696, null, null, f, null).
```

</details>

rousseldenis:
@JordiMForgeFlow `product_variant_name` module should not add a required field on existing model.

Don't you have another mean (only on forms) ?

marielejeune:
Hi @StefanRijnhart 
Even if I agree with Denis (to me we should avoid adding required fields without a default value if no logic asks it),
I've set `product_variant_name` as a rebel module so that we can merge this PR (which I rebased on 18.0).

StefanRijnhart:
Yes, agreed that this is a more viable strategy in the long turn. Thanks for implementing the workaround!

/ocabot merge nobump

PR #412, opened 2025-08-23
[17.0][MIG] product_variant_default_code: Migration to 17.0

pedrobaeza:
Reviving the migration in #341.

I have squashed some administrative commits and same-author translations, and reorder the fetched from 16 commit.

@Tecnativa TT52346

PR #413, opened 2025-08-23
[17.0][FIX] purchase_variant_configurator: _select_seller signature

pedrobaeza:
The signature of the method is not correct according upstream:

https://github.com/odoo/odoo/blob/8eddce363f05cc672c0c208239f86254f1175da1/addons/product/models/product_product.py#L682

so we should stick to the proper one.

@Tecnativa


pedrobaeza:
I'm fast-tracking for fixing the error in the branch manifested in #412.

/ocabot merge patch

PR #424, opened 2025-10-31
[19.0][MIG] sale_order_line_variant_description: Migration to 19.0

NICO-SOLUTIONS:
standard migration

While running the test, I noticed in my local setup that it fails if the website_sale module is installed. This happens because, in Odoo 19.0, website_sale adds a required field publish_date on the product.template model.

The module itself does not depend on website_sale, and its test only creates a standard product.template record (not in a website context). I assume this means the issue should be handled separately?! — e.g., in an extension module like website_sale_order_line_variant_description. 
Am I correct in assuming that the test case does not need to cover this combination, or should it be adapted to support the case when website_sale is installed?


NICO-SOLUTIONS:
@pedrobaeza 
Hey Pedro, could you please remove the stale tag when you have a moment? That would be great!

sale_order_line_variant_description/tests/test_sale_order_line_variant_description.py
```diff
@@ -0,0 +1,45 @@
+# Copyright 2015-17 Alex Comba - Agile Business Group
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo.tests import TransactionCase
+
+
+class TestSaleOrderLineVariantDescription(TransactionCase):
```
BhaveshHeliconia: Use BaseCommon class instead of TransactionCase because there is no need to define cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True)) and no need to create res.partner because in BaseCommon class already created so we can optimize code using BaseCommon.

BhaveshHeliconia (changes requested):

sale_order_line_variant_description/tests/test_sale_order_line_variant_description.py
NICO-SOLUTIONS: @BhaveshHeliconia 
Hey, yeah I’ll take care of that. The initial intention was just to migrate it here, but improving the test definitely makes sense

BhaveshHeliconia: Sounds good, thanks! Since this is a minor change, it could also be handled as part of the migration.

BhaveshHeliconia (approved):
LGTM!

ahaly-compass:
any update on this pull request?

NICO-SOLUTIONS:
> any update on this pull request?
> 
> 

@ahaly-compass 
Feel free to Review

NICO-SOLUTIONS:
@pedrobaeza 
sorry for the hassle... But could you remove the stale again?

NICO-SOLUTIONS:
> /ocabot migration sale_order_line_variant_description

Thx 🙏

ahaly-compass (approved):
Functions as expected

my2ndhead:
@OCA/product-maintainers
This PR is approved and ready to merge since Sep 17.
Could a maintainer run `/ocabot merge nobump`?

PR #425, opened 2025-11-10
[16.0] [IMP] product_variant_default_code : add sequence to complete …

emiliesoutiras:
…or replace patterns

When a product contains many attributes, it may be useful to simply use a prefix and a sequence.
This proposal consists of adding the possibility of defining two new patterns, one that will be replaced by a sequence value, global for specific to the product, and another to directly fill in the variant identifier.


product_variant_default_code/models/config_settings.py
```diff
@@ -23,3 +23,17 @@ class BaseConfiguration(models.TransientModel):
         default=False,
         config_parameter="product_variant_default_code.prefix_as_default_code",
     )
+
+    sequence_as_default_code = fields.Many2one(
+        string="Default sequence as default Reference",
+        comodel_name="ir.sequence",
+        default=False,
+        config_parameter="product_variant_default_code.seuence_as_default_code",
```
Kev-Roche:
```suggestion
        config_parameter="product_variant_default_code.sequence_as_default_code",
```

product_variant_default_code/models/config_settings.py
```diff
@@ -23,3 +23,17 @@ class BaseConfiguration(models.TransientModel):
         default=False,
         config_parameter="product_variant_default_code.prefix_as_default_code",
     )
+
+    sequence_as_default_code = fields.Many2one(
+        string="Default sequence as default Reference",
+        comodel_name="ir.sequence",
+        default=False,
```
Kev-Roche:
```suggestion
```

product_variant_default_code/models/config_settings.py
```diff
@@ -23,3 +23,17 @@ class BaseConfiguration(models.TransientModel):
...
+        config_parameter="product_variant_default_code.seuence_as_default_code",
+    )
+
+    use_sequence_per_product_tmp_as_default_code = fields.Boolean(
+        string="Use specific sequence par Product in Reference",
+        default=False,
```
Kev-Roche:
```suggestion
```

product_variant_default_code/views/config_settings_view.xml
```diff
@@ -30,6 +30,32 @@
...
+                            <field name="sequence_as_default_code" />
+                        </div>
+                        <div
+                            class="o_setting_right_pane"
+                            attrs="{'invisible': [('sequence_as_default_code', '=', True)]}"
+                        >
+                            <label for="sequence_as_default_code" />
+                            <div class="text-muted">
+                                Defines the "Reference Sequence" as the default for compute reference (default_code) of the variant.
+                            </div>
```
Kev-Roche, line 51: need to be fix in the view, we can not see the sequence name.

Kev-Roche (changes requested):

PR #433, opened 2026-01-30
[18.0][ADD] product_variant_search

AungKoKoLin1997:
Supersedes https://github.com/OCA/product-attribute/pull/2196
This module improves product variant searching by matching the product display name across all product selection fields.

@qrtl QT6354


product_variant_search/models/product_attribute_value.py
```diff
@@ -0,0 +1,21 @@
+# Copyright 2026 Quartile (https://www.quartile.co)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import models
+
+
+class ProductAttributeValue(models.Model):
+    _inherit = "product.attribute.value"
+
+    def _update_field_translations(self, field_name, translations, source_lang=None):
```
yostashiro: What is the reason we extend this method here, while we extend write() in product.template? Is there any problem doing the same (i.e. use write())?

product_variant_search/models/product_product.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+class ProductProduct(models.Model):
+    _inherit = "product.product"
+
+    variant_search_name = fields.Char(
+        readonly=True,
+        translate=True,
+        index=True,
+    )
```
yostashiro: Let's further remove the redundancy.
```suggestion
    search_name = fields.Char(readonly=True, translate=True, index=True)
```

product_variant_search/models/product_product.py
```diff
@@ -0,0 +1,54 @@
...
+        variants.assign_variant_search_name_all_langs()
+        return variants
+
+    def write(self, vals):
+        res = super().write(vals)
+        if "default_code" in vals:
+            self.assign_variant_search_name_all_langs()
+        return res
+
+    def assign_variant_search_name_all_langs(self):
```
yostashiro:
```suggestion
    def assign_search_name_all_langs(self):
```

product_variant_search/models/product_product.py
```diff
@@ -0,0 +1,54 @@
...
+        langs = self.env["res.lang"].search([("active", "=", True)]).mapped("code")
+        for lang in langs:
+            variants_lang = self.with_context(lang=lang)
+            for v in variants_lang:
+                v.write({"variant_search_name": v.display_name})
```
yostashiro: Can we pre-construct a jsonb value per product (covering all langs) and then write with raw SQL to improve the performance?

product_variant_search/models/product_product.py
```diff
@@ -0,0 +1,54 @@
...
+        limit_rest = (limit - len(res)) if limit else None
+        extra_domain = expression.AND(
+            [
+                args,
+                [("id", "not in", found_ids)],
+                [("variant_search_name", operator, name)],
+            ]
+        )
+        extra = self.search_fetch(extra_domain, ["display_name"], limit=limit_rest)
+        return res + [(p.id, p.display_name) for p in extra.sudo()]
```
yostashiro, line 66: Can you check if this works?
```suggestion
    @property
    def _rec_names_search(self):
        return super()._rec_names_search + ["search_name"]
```

AungKoKoLin1997: It doesn't work.

product_variant_search/models/product_attribute_value.py
```diff
@@ -0,0 +1,19 @@
...
+class ProductAttributeValue(models.Model):
+    _inherit = "product.attribute.value"
+
+    def write(self, vals):
+        res = super().write(vals)
+        if "name" in vals:
+            ptavs = self.env["product.template.attribute.value"].search(
+                [
+                    ("product_attribute_value_id", "in", self.ids),
+                ]
```
yostashiro:
```suggestion
                [("product_attribute_value_id", "in", self.ids)]
```

product_variant_search/hooks.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2026 Quartile (https://www.quartile.co)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+BATCH_SIZE = 2000
+
+
+def post_init_hook(env):
```
yostashiro, line 8: I was thinking how it would work for a database with millions of products. Can we add a cron to populate search_name? The cron can disable itself once all records have been updated.

This hook can still exist but we may set a hard cap to the number of products to process.

AungKoKoLin1997: @yostashiro I just realized we can't disable the cron itself because of `_try_lock()` is blocked the operation and raise error.

```
    raise UserError(_("Record cannot be modified right now: "
odoo.exceptions.UserError: Record cannot be modified right now: This cron task is currently being executed and may not be modified Please try again in a few minutes
```

AungKoKoLin1997: I tried using two crons: an `Assign Search Name` cron and a separate `Disable` cron set to `active=False`.
However, Odoo won’t execute inactive crons even if they have an `ir.cron.trigger` (the scheduler only picks
`active=true` jobs), so the disable cron never runs. We could bypass this by changing core scheduling, but
that may cause security/maintenance risk and isn’t worth it.
https://github.com/odoo/odoo/blob/ba0b3118bc0a628989c25e2159e3614c76847a0e/odoo/addons/base/models/ir_cron.py#L212-L228

Therefore, I decided to use a single cron and note in
`ROADMAP.md` that it needs to be disabled manually once all products have `search_name` assigned.

product_variant_search/data/ir_cron.xml
```diff
@@ -0,0 +1,16 @@
...
+        <field name="state">code</field>
+        <field name="code">model._cron_populate_search_name()</field>
+        <field name="interval_number">1</field>
+        <field name="interval_type">weeks</field>
+        <!-- start 5 minutes later -->
+        <field
+            name="nextcall"
+            eval="(DateTime.now() + relativedelta(minutes=5)).strftime('%Y-%m-%d %H:%M:%S')"
+        />
+        <field name="active">True</field>
```
yostashiro: Shouldn't this be disabled by default?

yostashiro: I think the cron should be disabled by default. The admin user should enable this as needed.
```suggestion
        <field name="active">False</field>
```

yostashiro (approved):
LGTM. 👍

product_variant_search/readme/DESCRIPTION.md
```diff
@@ -0,0 +1,5 @@
+This module improves product variant searching by matching the product display name
+across all product selection fields.
```
yostashiro: @AungKoKoLin1997 Can you please add more information? I think this description is a bit too vague.

product_variant_search/readme/INSTALL.md
yostashiro, line 1: Please switch this to INSTALL.md.

nobuQuartile (approved):
Code review and functional review: LGTM

I have confirmed that in a multi-language environment, product variants can be searched using attribute values in each respective language.

I have also confirmed that when the user’s language is set to English, searching with Japanese attribute values does not return the corresponding product variants in the search results.

product_variant_search/readme/DESCRIPTION.md
AungKoKoLin1997: Updated!

product_variant_search/readme/INSTALL.md
AungKoKoLin1997: Done!

smorita7749 (approved):
Functional review: LGTM
Create a product with the following fields:

Product name: Desk  
Attribute: Color  
Values: Wood, Black  

When searching for "Wood" in the product section field, the product variant "Desk (Wood)" is displayed.
When searching for "Desk (Black)" in the product section field, the targeted product variant is displayed.

PR #437, opened 2026-03-17
[18.0][IMP] product_variant_configurator_manual_creation: improve per…

astirpe:
…formance


astirpe:
@ChrisOForgeFlow, do you agree with this proposed performance improvement? We are already using it in production.

BhaveshHeliconia (approved):
LGTM!

PR #443, opened 2026-05-12
[19.0][MIG] sale_variant_configurator: Migration to 19.0

Andrii9090-tecnativa:
Changes:
 - `_action_confirm` -> `action_confirm`: this is needed to avoid the `UserError` on [sale.order](https://github.com/OCA/OCB/blob/19.0/addons/sale/models/sale_order.py#L1203)
- Disabled the `required` attribute for `product_id`, `product_uom_id`, and `product_template_id` in `sale.view_order_form`.

image

@pedrobaeza ping
@Tecnativa
TT61966


sale_variant_configurator/readme/CONTRIBUTORS.md
```diff
@@ -1,2 +1,6 @@
+- Oihane Crucelaegui <oihanecrucelaegi@avanzosc.es>
+- Pedro M. Baeza <pedro.baeza@tecnativa.com>
```
pedrobaeza: Don't put emails and group our references by company (and include you).

sale_variant_configurator/views/sale_view.xml
```diff
@@ -54,7 +54,7 @@
                     <list create="0" delete="0" editable="bottom">
                         <field name="owner_model" column_invisible="1" />
                         <field name="owner_id" column_invisible="1" />
-                        <field name="attribute_id" column_invisible="1" />
+                        <field name="attribute_id" force_save="1" />
```
pedrobaeza, line 76: This should remain column invisible.

sale_variant_configurator/views/sale_view.xml
```diff
@@ -69,6 +69,39 @@
...
+                expr="//field[@name='order_line']/list//field[@name='product_uom_id']"
+                position="attributes"
+            >
+                <attribute name="required">0</attribute>
+            </xpath>
+            <xpath
+                expr="//field[@name='order_line']/list//field[@name='product_template_id']"
+                position="attributes"
+            >
+                <attribute name="required">0</attribute>
```
pedrobaeza: Why this one is not required?

pedrobaeza (changes requested):
Apart from the comments inline, we should keep only one column for showing "Product"/"Product template", and more as modifications are handled in the popup.

More problems detected:

- A variant without attributes have been created on sales confirmation:

Image

- The text with the attribute values have been lost (probably because the final selected variant was the empty one):

Image

sale_variant_configurator/views/sale_view.xml
Andrii9090-tecnativa: If not, we don't save the form 

Image

victoralmau:
Please, cherry-pick https://github.com/OCA/product-variant/pull/449 to commit history before migration commit

victoralmau:
Please, cherry-pick https://github.com/OCA/product-variant/pull/448 to commit history before migration commit.

sale_variant_configurator/models/sale_order.py
```diff
@@ -9,14 +9,17 @@
...
+    def action_confirm(self):
         """Create possible product variants not yet created."""
         lines_without_product = self.mapped("order_line").filtered(
             lambda x: not x.product_id and x.product_tmpl_id
         )
         for line in lines_without_product:
-            line.create_variant_if_needed()
-        return super()._action_confirm()
+            if bool(line.product_attribute_ids.mapped("value_id")):
+                line.create_variant_if_needed()
```
victoralmau: I don't understand the reason for that condition; I think the `create_variant_if_needed()` method should always be called (without any conditions).

sale_variant_configurator/models/sale_order.py
```diff
@@ -9,14 +9,17 @@
...
+            else:
+                line.product_template_id = line.product_tmpl_id
```
victoralmau: I don't understand this; I don't think it's necessary. If something needs to be defined (for example, the `product_template_id` field), it should be done in the `create_variant_if_needed()` method.

sale_variant_configurator/models/sale_order.py
```diff
@@ -30,25 +33,19 @@ class SaleOrderLine(models.Model):
         related=False,
         string="Product Template (no related)",
     )
-    product_id = fields.Many2one(required=False)
```
victoralmau: I found it odd to remove this, but I can see that the field has never been set to `required=True`.

victoralmau:
Please, cherry-pick https://github.com/OCA/product-variant/pull/450 to commit history before migration commit and remove unnecessary changes in the migration commit (related to https://github.com/OCA/product-variant/pull/443#discussion_r3243084247)

sale_variant_configurator/views/sale_view.xml
```diff
@@ -88,6 +88,23 @@
...
+                >not display_type and not is_downpayment and not product_tmpl_id</attribute>
+            </xpath>
+            <xpath
+                expr="//field[@name='order_line']/form//field[@name='product_uom_id']"
+                position="attributes"
+            >
+                <attribute
+                    name="required"
+                >not display_type and not is_downpayment and not product_tmpl_id</attribute>
+            </xpath>
```
victoralmau: That isn't necessary, according to https://github.com/OCA/product-variant/pull/450

victoralmau (changes requested):

Andrii9090-tecnativa:
@victoralmau  you are right. I changed the code, tested it, and it works.

sale_variant_configurator/models/sale_order.py
```diff
@@ -105,3 +99,13 @@ def _compute_product_uom(self):
...
+
+    def _get_sale_order_line_multiline_description_variants(self):
+        # It needs to correctly show the name of the sale.order.line with attributes
+        name = super()._get_sale_order_line_multiline_description_variants()
+        if not self.product_attribute_ids:
+            return name
+        for attribute in self.product_attribute_ids:
+            if attribute.value_id.display_name:
+                name += "\n" + attribute.value_id.display_name
+        return name
```
victoralmau: Why is this change necessary? The `_get_sale_order_line_multiline_description_variants()` method already existed in v18 and v16; if it’s necessary, it will also be necessary in those versions first. 
The appropriate change would be something like:
```

if self.product_tmpl_id and not self.product:
     name = xxx
else:
   name = super()._get_sale_order_line_multiline_description_variants()
return name
```

It would also be a good idea to add a test to verify the relevant use case.

I've tested it in v19 without this overridden method, and it works fine, in my opinion.

Image

victoralmau (approved):

pedrobaeza (approved):
/ocabot merge nobump

PR #446, opened 2026-05-21
[16.0][FIX] product_variant_configurator: prevent newId wrapping of existing template on new variants

JordiMForgeFlow:
This issue was noticed because some products were missing the values of _property_stock_production_ and _property_stock_inventory_ when there are default company properties that set their values.

The issue is reproduced like:

1. Create a new template. The company dependent fields should be filled in based on the default company properties (you can use for example _property_stock_production_ or _property_stock_inventory_).
2. From the product variants menu, create a new variant and assign it to the previously created template. The moment you assign it (even before saving the new variant) you will see that the company dependent fields get cleared out.
3. When saving the new variant, the cleared out values are written and the product ends up without the values set.

After investigating, the issue is affecting company-dependent fields that are managed at the product template level. It actually also affects other models that follow the __inherits_ structure. The reason why this happens is because:

1. For m2o fields related to __inherits_ structure (delegated) Odoo wraps the value as a newId if we are creating a new record: https://github.com/odoo/odoo/blob/c415f984c8ba4e636801fce752eb6fd2c0b8082e/odoo/fields.py#L3011. Odoo basically assumes that, if you are creating a new variant in the Form, you are also creating a new template.
2. When the template is assigned, the fields are then computed in https://github.com/odoo/odoo/blob/c415f984c8ba4e636801fce752eb6fd2c0b8082e/odoo/fields.py#L749 but the template is evaluated as a NewId so the results are False.

A patch on the _convert_to_cache_ method solves the issue and seems to be the right place to fix. Nonetheless, I am interested in other opinions on this.

### DIFF
--- a/product_variant_configurator/__init__.py
+++ b/product_variant_configurator/__init__.py
@@ -1,1 +1,2 @@
+from . import hooks
 from . import models
--- a/product_variant_configurator/hooks.py
+++ b/product_variant_configurator/hooks.py
@@ -1,0 +1,42 @@
+# Copyright 2026 ForgeFlow S.L.
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+from odoo import fields
+from odoo.models import NewId
+
+
+def _patch_many2one_no_wrap_existing_parent():
+    """Patch ``Many2one.convert_to_cache`` so that linking a new child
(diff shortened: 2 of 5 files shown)


AaronHForgeFlow (approved):
Code review LGTM.

Nice catch!

PR #455, opened 2026-07-06
[19.0][FIX] sale_variant_configurator: Add product_tmpl_id.uom_ids to allowed_uom_ids (sale.order.line)

Andrii9090-tecnativa:
`sale.order.line.product_uom_id` has the domain` [("id", "in", allowed_uom_ids)]`. If `product_id` is not set, we can't select a UoM for this sale order line because `allowed_uom_ids` is empty.

Before fix
image

After fix
image

@pedrobaeza can you review, please?
@Tecnativa

### DIFF
--- a/sale_variant_configurator/models/sale_order.py
+++ b/sale_variant_configurator/models/sale_order.py
@@ -104,3 +104,13 @@ class SaleOrderLine(models.Model):
             line.product_uom = line.product_tmpl_id.uom_id
         self -= lines_with_template
         return super()._compute_product_uom()
+
+    @api.depends("product_tmpl_id")
+    def _compute_allowed_uom_ids(self):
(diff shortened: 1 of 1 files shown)


sale_variant_configurator/models/sale_order.py
```diff
@@ -104,3 +104,12 @@ def _compute_product_uom(self):
             line.product_uom = line.product_tmpl_id.uom_id
         self -= lines_with_template
         return super()._compute_product_uom()
+
+    def _compute_allowed_uom_ids(self):
```
pedrobaeza, line 109: Shouldn't this add `@api.depends("product_tmpl_id")`?

pedrobaeza (approved):
/ocabot merge patch

PR #458, opened 2026-07-13
[18.0][FIX] product_variant_default_code: init hooks, to avoid MemoryError

oihane:


sbiosca-s73:
@oihane Could you please also remove `self.env.cr.flush()` from `_compute_default_code` in version 19? The same issue that occurred in version 18 also affects version 19.

anajuaristi (approved):

anajuaristi:
Hi  @Kev-Roche, could this be merged please? 
It solves a memory error traceback and we need it to be used in production.
Thank you!

PR #460, opened 2026-07-22
[14.0][FIX] product_variant_inactive: archive template with active_test

Kev-Roche:
_compute_active tested the emptiness of the variant recordset instead of the active values. With active_test disabled,  roduct_variant_ids still returns the archived variants, so the template was set back to active.
This happens while archiving a template: the variants are archived through with_context(active_test=False), and a flush from that environment recomputes the field. Archiving then silently fails.
Use any() with an explicit active_test=False context.


rousseldenis (approved):

sebastienbeau (approved):
LGTM (code review)

rousseldenis:
@Kev-Roche Could you fix pre-commit ?

PR #462, opened 2026-08-20
[19.0][FIX] sale_variant_configurator: Apply the sale order partner’s language

Andrii9090-tecnativa:
Apply the sale order partner’s language when generating sale order line descriptions.
This ensures that sale order line descriptions are correctly translated.

@pedrobaeza can you review, please?

@Tecnativa
TT64253

PR #463, opened 2026-08-21
[19.0][FIX] product_variant_configurator: translate

Andrii9090-tecnativa:
Revert commit #462 and added the changes that were lost since the [v16](https://github.com/OCA/product-variant/commit/2351a7b22da9b22de461e71cbc004493e648a5f4#diff-da880cc7befc12d224f5853cd6b452dbbc02c1074d03c2019391c7630f1bb4aeR154) migration and are missing in the [v17](https://github.com/OCA/product-variant/commit/516d71ae291d182d35cad4db951f81233a569237) migration. These changes are essential to use the product name, variant names, and description in the partner’s language context.

`sale.order.line` doesn't have a `partner_id` field; it uses [order_partner_id](https://github.com/odoo/odoo/blob/19.0/addons/sale/models/sale_order_line.py#L48). Therefore, without these changes, the partner’s language context is not applied. 
`SaleOrderLine` has [_partner_id_field](https://github.com/OCA/product-variant/blob/19.0/sale_variant_configurator/models/sale_order.py#L25) defined.

@pedrobaeza @victoralmau can you review?

TT64264
@Tecnativa

### DIFF
--- a/sale_variant_configurator/models/sale_order.py
+++ b/sale_variant_configurator/models/sale_order.py
@@ -68,8 +68,7 @@ class SaleOrderLine(models.Model):
         res = super()._get_product_description(
             template=template, product=product, product_attributes=product_attributes
         )
-        product = self.product_id.with_context(lang=self.order_partner_id.lang)
-        if product.description_sale:
+        if product and product.description_sale:
             res = (res or "") + "\n" + product.description_sale
         return res
 
--- a/sale_variant_configurator/models/sale_order.py
+++ b/sale_variant_configurator/models/sale_order.py
@@ -65,16 +65,11 @@ class SaleOrderLine(models.Model):
 
     @api.model
     def _get_product_description(self, template, product, product_attributes):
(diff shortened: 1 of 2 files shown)


sale_variant_configurator/models/sale_order.py
```diff
@@ -65,16 +65,11 @@ def create(self, vals_list):
...
-        if lang != self.env.lang:
-            if template:
-                template = template.with_context(lang=lang)
-            if product:
-                product = product.with_context(lang=lang)
         res = super()._get_product_description(
             template=template, product=product, product_attributes=product_attributes
         )
-        if product and product.description_sale:
+        product = self.product_id.with_context(lang=self.order_partner_id.lang)
```
pedrobaeza: We can remove this part as well and let `if product and product.description_sale`, as the calling method already puts the lang env.

pedrobaeza (approved):

victoralmau (approved):

PR #464, opened 2026-08-25
[19.0][FIX] product_variant_configurator: Create product with sudo

pedrobaeza:
The user confirming the sales/purchase may not have permission to create products, but it's logical to not require such permission in these flows, as this is only creating a variant of an existing template, so let's make it with sudo.

@Tecnativa TT64287

### DIFF
--- a/product_variant_configurator/models/product_configurator.py
+++ b/product_variant_configurator/models/product_configurator.py
@@ -302,7 +302,9 @@ class ProductConfigurator(models.AbstractModel):
                     prod_attr_val=product_attribute_value: v.product_attribute_value_id  # noqa
                     == prod_attr_val
                 )
-            product = product_obj.create(
+            # Create the product with sudo for not requiring the permission for creating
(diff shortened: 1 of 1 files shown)


CarlosRoca13 (approved):

product_variant_configurator/models/product_configurator.py
```diff
@@ -302,7 +302,7 @@ def create_variant_if_needed(self):
                     prod_attr_val=product_attribute_value: v.product_attribute_value_id  # noqa
                     == prod_attr_val
                 )
-            product = product_obj.create(
+            product = product_obj.sudo().create(
```
victoralmau, line 307: Perhaps a comment explaining why this is necessary might be helpful in the future.

victoralmau (approved):

product_variant_configurator/models/product_configurator.py
pedrobaeza: Totally, done!

